PR #220, opened 2014-03-16
"Build Menu"-event

KingCrunch:
I've seen this and I wonder: Is theres a reason, why this isn't built-in?

The way the event is trigger looks quite generic

```php
$this->container->get('event_dispatcher')->dispatch(ConfigureMenuEvent::CONFIGURE, new ConfigureMenuEvent($factory, $menu));
```

So after (in this case) `build()` was called from [BuilderAliasProvider](https://github.com/KnpLabs/KnpMenuBundle/blob/master/Provider/BuilderAliasProvider.php#L64) the same class could trigger the event with the return value and the factory, that was previously passed as first argument of that method.


KingCrunch:
Ping

PR #243, opened 2014-11-26
[RFC] Define menus in config, use the ExpressionLanguage for labels and show/hide logic

kbond:
This PR allow you to define menus in the bundle config:

``` yaml
knp_menu:
    menus:
        my_menu:
            children:
                login:
                    route: fos_user_security_login
                logout:
                    route: fos_user_security_logout
```

The menu can be rendered in your templates with `knp_menu_render('my_menu')`

Additionally, you can use the expressions to dynamically change the label or show/hide items:

``` yaml
knp_menu:
    menus:
        my_menu:
            children:
                login:
                    route: fos_user_security_login
                    hide_if: "@=is_granted('ROLE_USER')"
                logout:
                    label: "@='Logout '~user"
                    route: fos_user_security_logout
                    show_if: "@=is_granted('ROLE_USER')"
```
- Expressions prefixed with `@=`
- `hide_if`/`show_if` do not require the `@=` but probably is a good practice
- Variables available in expressions: `user`, `security`, `request`
- Functions available in expressions: `is_granted`

You can use expressions for any item:

``` yaml
knp_menu:
    menus:
        my_menu:
            children:
                login:
                    route: fos_user_security_login
                    hide_if: "@=is_granted('ROLE_USER')"
                logout:
                    route: fos_user_security_logout
                    show_if: "@=is_granted('ROLE_USER')"
                    extras:
                        translation_params:
                            "%%user%%": @=user
```

TODO:
- Docs
- <del>Tests</del>
- Improve the configuration to better validate the menus(?)
- <del>Optimize expressions during container compilation. I am not an expert on the ExpressionLanguage component. Maybe someone else can help with this? I would like to not have the `expression` and `extra.translation_params_expression` options and just do `label: @=expression` but symfony throws and error when using `@=` in parameters...</del> I think these are optimized now.
- <del>Ability to extend and customize the expression language and expression evaluator</del>

If this is something you would want in the bundle, I will write the documentation.  This PR is sort of two parts: the config provider and the expression loader/extension. If you'd like, I can split this PR.


stof:
I don't like all the logic you add around the ExpressionLanguage, and even less the fact that you provide your own way to provide functions for the ExpressionLanguage rather than using features available in Symfony (and the way you use here has been rejected in Symfony because it mixes concerns in the same class and creates issues).

phpunit.xml.dist
```diff
@@ -1,6 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <phpunit colors="false" bootstrap="vendor/autoload.php">
+    <php>
+        <!-- Disable E_USER_DEPRECATED until 3.0 -->
+        <!-- php -r 'echo -1 & ~E_USER_DEPRECATED;' -->
+        <ini name="error_reporting" value="-16385"/>
```
dbu: this would have to use the error reporting helper from symfony instead.

stof: we use it now. But this helper was not yet implemented in Symfony at the time this PR was sent.

dbu:
@stof so would you propose to just keep the static menu definition of this PR and drop expression language for this? that would make sense to me. expression language leads to people writing EL code instead of PHP code. The PHP code would be more readable and not that huge.

defining static menus instead of writing a twig fragment that has both the structure *and* the HTML formatting would make sense to me however.

to throw in a different idea: what about adding menu information on the routing definition? similar to what this bundle is doing: https://github.com/thormeier/breadcrumb-bundle#basic
we would have to scan the route collection when building the cache. but then it would be simple for the user to make sure their menu entries actually match routes that exist.

dbu:
how much does https://github.com/KnpLabs/KnpMenuBundle/pull/273 already simplify what you try to do here? (minus the expression language that we don't want to put into the bundle).

phpunit.xml.dist
dbu: right. so revert the change and rebase on master. (which is needed anyways, there is a conflict reported by github)

stof:
@dbu I don't like the idea of putting this info in the routing files. It would lead to a complex setting (to avoid bypassing the router cache, which the bundle you linked fails to do) and would clutter the routing file with lots of menu-related info.
Thus, it would make things very hard to understand: the same route can appear in multiple menus, and we would have no way to see easily what a menu contains (a breadcrumb configured in the routing configuration directly makes more sense than a full menu)

DependencyInjection/Configuration.php
```diff
@@ -41,6 +41,13 @@ public function getConfigTreeBuilder()
                 ->end()
                 ->booleanNode('templating')->defaultFalse()->end()
                 ->scalarNode('default_renderer')->cannotBeEmpty()->defaultValue('twig')->end()
+                ->variableNode('menus')
+                    ->defaultNull()
+                    ->validate()
+                    ->ifTrue(function ($value) { return !is_array($value); })
```
stof: then it should not be a variable node, but a prototyped array node

DependencyInjection/KnpMenuExtension.php
```diff
@@ -38,10 +40,45 @@ public function load(array $configs, ContainerBuilder $container)
...
     }
 
     public function getNamespace()
     {
         return 'http://knplabs.com/schema/dic/menu';
     }
+
+    private function preParseExpressions(array $menus, ContainerBuilder $container)
+    {
+        $contextClass = $container->getParameter('knp_menu.expression_context.class');
```
stof: This will always be the same class here, because nothing has a chance to change the parameter between the loading of the file and the call to ``preParseExpressions``. You are making the code more complex for nothing.

stof:
btw, having a huge blob on unvalidated (and undocumented) configuration might not be the best idea.

kbond:
To be honest, I am not seeing the benefit of this PR without the expressions. To create a static menu, just pass some yaml config to a menu that's a service.

dbu:
@stof should we close this or do you think we will find a solution that can find a place in this bundle?

stof:
Well, I'm not comfortable with adding such feature in KnpMenuBundle itself (partially because I don't like the fact of putting so much business logic in a config file using expressions).
I suggest you to build this as a separate bundle (depending on KnpMenuBundle), which would then register a MenuProvider for the menus it is building (or register builders for these menus in the existing MenuProvider of KnpMenuBundle depending on what is easier for you)

stof:
I'm closing this PR. For the reasoning, see my previous comments

PR #249, opened 2014-12-15
Proposed update to 2.6 "Best Practices" plus..

DougHayward:
Proposed update to 2.6 "Best Practices" plus added Sub Menu description.

### DIFF
--- a/Resources/doc/index.md
+++ b/Resources/doc/index.md
@@ -95,8 +95,8 @@ An example builder class would look like this:
 
 ```php
 <?php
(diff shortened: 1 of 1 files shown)


javiereguiluz:
:+1: I love this change. Thanks @BonnieDoug.

In case KnpMenuBundle maintainers aren't convinced yet about this change, all the official Symfony docs have been switched to AppBundle. Moreover, using `Acme/DemoBundle` is strongly discouraged because showing a "demo bundle" looks like non-professional application development.

PR #254, opened 2015-01-19
fix doc link

mhor:


Nek-:
:+1: nice catch ^^

stof:
Cna you switch it to a relative link (using ``Resources/doc/index.rst``) so that github makes it point to the same version than the one the readme is displayed for ? It will be even better.

@javiereguiluz is there any ETA for having it rendered on symfony.com ?

mhor:
@stof done :+1:

stof:
thanks

javiereguiluz:
@stof I don't know when will the bundles' doc be integrated in symfony.com, but I guess it will be integrated soon because everything is working OK locally:

knpmenubundle_in_symfony_com

stof:
@javiereguiluz it would be great to have it soon, because configuration blocks don't render on github, so this means we need to have a rendered doc somewhere else. If it does not come on symfony.com soon enough, this would require doing a RTD setup to avoid breaking the DX. Who would know about this ETA ?

Btw, I see a bug in your screenshot: the URL says ``master`` but the version switcher says ``1.1``. the master branch of KnpMenuBundle is not 1.1. It is 2.0 (and the 1.1 doc has not been converted to rST)

PR #255, opened 2015-01-19
Fixed some minor syntax issues in the docs

javiereguiluz:
I've integrated the documentation of this bundle in my local version of symfony.com website and I've found some syntax issues. This Pull Requests fixes those issues and hopefully is the last one related to docs formatting.

### DIFF
--- a/Resources/doc/events.rst
+++ b/Resources/doc/events.rst
@@ -2,7 +2,7 @@ Using events to allow a menu to be extended
 ===========================================
 
 If you want to let different parts of your system hook into the building of your
-menu, a good way is to use an approach based on the Symfony2 EventDispatcher
+menu, a good way is to use an approach based on the Symfony EventDispatcher
 component.
 
 Create the menu builder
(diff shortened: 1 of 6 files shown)


Resources/doc/events.rst
```diff
@@ -50,54 +50,54 @@ Create the Event object
...
 
-      /**
-       * @return \Knp\Menu\ItemInterface
-       */
-      public function getMenu()
-      {
-          return $this->menu;
-      }
-  }
+    <?php
```
stof: AFAIK, the open tag needs to be omitted for PHP code blocks, because the sphinx config registers the Pygments PHP lexer with ``startInline=True``

Resources/doc/events.rst
```diff
@@ -50,54 +50,54 @@ Create the Event object
...
+    // src/Acme/DemoBundle/Event/ConfigureMenuEvent.php
+
+    namespace Acme\DemoBundle\Event;
```
stof: I think it would be worth migrating from AcmeDemoBundle to AppBundle to match the Symfony best practices

Resources/doc/events.rst
stof: AFAIK, the open tag needs to be omitted for PHP code blocks, because the sphinx config registers the Pygments PHP lexer with ``startInline=True``

Resources/doc/events.rst
stof: In some other pages, we are already using AppBundle btw

javiereguiluz: Good idea. I've just updated all the docs to use the AppBundle where appropriate. Thanks.

stof:
Thank you @javiereguiluz.

PR #257, opened 2015-01-27
Fixed Rst syntax and added configuration blocks

WouterJ:


Resources/doc/index.rst
```diff
@@ -14,15 +14,15 @@ following command to download the latest stable version of this bundle:
 
 .. code-block:: bash
 
-    $ composer require knplabs/knp-menu-bundle
+    $ composer require knplabs/knp-menu-bundle "~2"
```
stof: not needed as Composer already does it

WouterJ: Hmm, I think I'll have to explain this to all bundle maintainers :)

The reason I added it is because you're documenting a specific version. Composer will just select te most recent version (if possible). So imagine you release KnpMenuBundle 3.0. In that case, the 2.0 docs would give you installation instructions for KnpMenuBundle 3.0 (as the version is omitted and composer just picks the latests).

stof:
Most syntax fixes have already been fixed in #255 but the configuration blocks are appreciated. Thanks @WouterJ.

PR #259, opened 2015-01-27
Replace Request with RequestStack in docs

bocharsky-bw:
Since Symfony 2.4 better to use [RequestStack](http://symfony.com/blog/new-in-symfony-2-4-the-request-stack) instead of Request.


stof:
Thanks @bocharsky-bw.

Schyzophrenic:
This was merged in the 2.0 branch of the documentation, whereas RequestStack object does not exist in 2.3.
The documentation is marked as "2.0" on SF website.
Maybe we should make a branch here.
@stof How is the documentation maintained for this ?

PR #263, opened 2015-03-16
'About me' child does not exist

jirikrepl:
current code results in error
Error: Call to a member function addChild() on null

so I added:
$menu->addChild('About Me', array('route' => 'about'));
and then you can add sub menu item "edit profile"


stof:
thanks

PR #268, opened 2015-05-05
Add priority to the knp_menu.matcher::addVoter

FlashBlack:
And piece of code in knp-menu
see: https://github.com/KnpLabs/KnpMenu/pull/197


stof:
this should be implemented without any change to KnpMenu actually

Nek-:
@Einenlum this PR could be close. (IMO the one on KnpMenu can be too)

dbu:
there is now #281

PR #271, opened 2015-06-15
Fixed version number in the documentation.

Nek-:
`~2` can be a problem if we break the compatibility in ~3 (and we will)


stof:
this is not. Semver-based operators in Composer never allow to get the next major version. ``~2`` and ``~2.0`` are equivalent

WouterJ:
Closing as this PR did not change anything, as explained by @stof

PR #272, opened 2015-06-16
Menu Builder as services

hshn:
Hi,

I would like to inject several services to a menu builder to build a menu.
I know that this bundle already provides the way to build menu and make it a service,
but I think the way is unable to retrieve builder options.

This PR provides the way to retrieve builder options and to inject services to the builder.

thanks


stof:
I actually have such a feature in my work project since a very long time and I wanted to move it to KnpMenuBundle itself. My implementation is a bit different and I tend to prefer the way it works. I will open a PR submitting it.

stof:
I just sent a PR with my implementation. you can have a look at it to see how it works.

hshn:
@stof 

Thank you for your quick reply and reaction.

I looked #273, and I felt that it's great to use yours!
So, I'm closing this PR.

PR #273, opened 2015-06-16
Add a menu provider registering builders as services

stof:
This is an alternative to #272

This code is running in production at Incenteev since 2 years (in a different namespace of course) and I finally took the time to extract it to KnpMenuBundle.
I wrote the documentation about the feature, so you can look at it to see how it works.


Nek-:
Looks good :) . :+1:

PR #278, opened 2015-08-27
Add ability to configure default menu options

WouterJ:
In the CMF, menu's aren't created programatically. With the current code, it means the menu options needs to be saved in the database in order to be set on the menu items.

It would be much better if we could configure this in the application settings. This PR implements a very simple DefaultOptionsExtensions, setting the default options based on a configured set of options.

This extension is automatically registered as soon as at least one option is configured.

@stof @Nek- I would love to hear your opinion about this (both the idea and whether it should be in this bundle or in the CMF).

Related issue: https://github.com/symfony-cmf/MenuBundle/issues/128


Extension/DefaultOptionsExtension.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace Knp\Bundle\MenuBundle\Extension;
+
+use Knp\Menu\Factory\ExtensionInterface;
+use Knp\Menu\ItemInterface;
+
+/**
+ * An extension to dynamically configure default menu options.
```
dbu: why "dynamically"? if the default options are in the configuration file, i would consider that rather static.

Resources/doc/menu_options.rst
```diff
@@ -0,0 +1,46 @@
+Setting Default Menu Options
+============================
+
+When creating a menu, menu options are set by passing them to ``addChild()``.
+When all menu items need the same option, adding the same options to each child
+can be cumbersome. In some cases, passing options isn't possible (e.g. when the
+menu was created based on a database).
```
dbu: i think the last line is not needed - and a bit confusing. i can store the options in the database and have them passed.

i think the point is really avoiding redundancy or having the effort to carry around default values in your application (that you probably set up in configuration anyways). in the cmf effort its the redundancy of storing options with every node which is an even worse redundancy.

dbu:
i like it!

Resources/config/options_extension.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <parameters>
+        <parameter key="knp_menu.extension.default_options.class">Knp\Bundle\MenuBundle\Extension\DefaultOptionsExtension</parameter>
```
stof: please don't introduce new class parameters. this practice is now discouraged in symfony

WouterJ: Ok, in the CMF, we decided to add them in the 1.x serie, for consistency and remove them all in 2.0. So I was a bit unsure about the standard of KnpMenu :)

WouterJ:
updated

dbu:
:+1: 

@Nek- do you agree with this one?

DependencyInjection/KnpMenuExtension.php
```diff
@@ -38,6 +38,11 @@ public function load(array $configs, ContainerBuilder $container)
         }
 
         $container->setParameter('knp_menu.default_renderer', $config['default_renderer']);
+
+        if (isset($config['default_menu_options'])) {
```
stof: it is always set (prototyped array nodes have a default value of empty array)

dbu:
oh, right. otherwise do you agree with the PR @stof or is that the only issue that needs to be fixed?

Extension/DefaultOptionsExtension.php
```diff
@@ -0,0 +1,39 @@
...
+    {
+        $this->options = $options;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function buildOptions(array $options = array())
+    {
+        return array_merge($this->options, $options);
```
stof: many options of the library are arrays. This makes them hard to use with your extension

Resources/doc/menu_options.rst
```diff
@@ -0,0 +1,45 @@
...
+In these cases, you can configure the menu options in the configuration file:
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # app/config/config.yml
+        knp_menu:
+            default_menu_options:
+                class: menu-item
```
stof: there is no ``class`` option in KnpMenu. Please use a real use case in the documentation (which will also help you see that the current implementation is not very usage)

Extension/DefaultOptionsExtension.php
WouterJ: So you suggest using `array_merge_recursive` instead?

stof: I don't know what to use here. ``array_merge_recursive`` has a very weird behavior when associate arrays are involved, so it is not the solution either (it would be even worse). ``array_replace_recursive`` would already be a bit better as we mostly have associative arrays (but this one leads to an unexpected behavior when lists are merged together though, and this happens in extras, for instance for the list of matched routes)

dbu:
this was in the 2.1 milestone but @stof just tagged 2.1.0

should we remove the milestone now? does it go to 2.2?

stof:
Well, I don't rely on the milestone in the project actually. I won't delay a release because a new feature is not ready. It can simply go in a new release.

garak:
@wouterj is this PR still active?
Some comments need to be addressed, and some conflicts need to be solved

PR #280, opened 2015-08-29
Translate menu items by default

### LINKED ISSUE #252: Suggest making the trans filter standard instead of having to extend the knp_menu.html.twig
Given that all apps these days should be prepared for translation from the get go, I'm wondering if it would be any idea to simplify the translation process of menu items by adding the trans filter to the default knp_menu.html.twig instead of the developer having to extend the default template just to add translation support? Or is there a technical/design reason it's not part of the default template? Afaik there really is no performance penalty since the menu is cached in production anyway?

WouterJ:
Fixes #252

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -37,7 +37,7 @@ class Configuration implements ConfigurationInterface
                     ->addDefaultsIfNotSet()
                     ->canBeUnset()
                     ->children()
-                        ->scalarNode('template')->defaultValue('knp_menu.html.twig')->end()
+                        ->scalarNode('template')->defaultValue('KnpMenuBundle::menu.html.twig')->end()
                     ->end()
                 ->end()
                 ->booleanNode('templating')->defaultFalse()->end()
--- a/Resources/doc/i18n.rst
+++ b/Resources/doc/i18n.rst
@@ -1,74 +1,64 @@
-I18n for your menu labels
+I18n for your Menu Labels
 =========================
 
-Let's assume you already have a working translation system.
-
-Create the template
--------------------
-
-At first, you need to create a template which extends ``knp_menu.html.twig``
-and adds the Twig ``trans`` filter. In this example, the template is located
-at ``src/AppBundle/Resources/views/Default/knp_menu.html.twig``. It allows changing
-the domain and the parameters for the translator through the extras of the item.
-
-.. code-block:: html+jinja
-
-    {% extends 'knp_menu.html.twig' %}
-    {% block label %}
-        {{ item.label|trans(
-            item.getExtra('translation_params', {}),
-            item.getExtra('translation_domain', 'messages')
-        ) }}
-    {% endblock %}
-
-Include the template
---------------------
-
-Alternative A: I18n for all menus
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-If you want to translate all of your menus, change the default template:
-
-.. code-block:: yaml
-
-    # app/config/config.yml
-    knp_menu:
-        twig:
-            template: AppBundle:Default:knp_menu.html.twig
-
-Alternative B: I18n only for certain menus
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-If you want to translate a certain menu, add the ``template`` parameter to
-its ``knp_menu_render`` call.
-
-.. code-block:: html+jinja
-
-    {{ knp_menu_render('name_of_your_menu', {'template': 'AppBundle:Default:knp_menu.html.twig'}) }}
-
-That's it.
-
-Add translation domain information in the controller
-----------------------------------------------------
-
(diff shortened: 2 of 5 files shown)


Resources/doc/i18n.rst
```diff
@@ -1,74 +1,58 @@
...
-    // will look for "Home" in AppBundle/Resources/translations/AppBundle.locale.yml
-    $menu->addChild('Home', array('route' => 'homepage'))->setExtra('translation_domain', 'AppBundle');
+You can configure the translation domain that's used in the extras of the menu
+item::
 
-    // will look for "Login" in Acme/AdminBundle/Resources/translations/AcmeAdminBundle.locale.yml
-    $menu->addChild('Login', array('route' => 'login'))->setExtra('translation_domain', 'AcmeAdminBundle');
+    // ...
+    $menu->addChild('Home', array('route' => 'homepage'))
+        ->setExtra('translation_domain' => 'AcmeAdminBundle');
```
dbu: if we add the menu defaults, this domain could also be configured via the configuration, right? and should it be AppBundle rather than AcmeAdminBundle?

dbu:
:+1: 

there is a configuration test that needs to be adapted.

Resources/doc/i18n.rst
WouterJ: Yes, the extra option can be used to configure extra's of an item.

I don't want people to use `AppBundle`, as people should use the default `messages` domain for their app translations imo. Custom domains are usefull for shared bundles.

dbu: makes sense

Resources/views/menu.html.twig
```diff
@@ -0,0 +1,3 @@
+{% extends 'knp_menu.html.twig' %}
+
+{% block label %}{{ parent()|trans({}, item.getExtra('translation_domain', 'messages')) }}{% endblock %}
```
stof: this is wrong:

- It breaks the support of safe labels available in the library, because it will always escape the return value of the trans tag
- it will double-escape the existing label (in case the translator does not change the string) because it is already escaped in the parent
- the translator might fail to translate if the escaping in the parent changes the string as it won't match the translation key anymore.

dbu: so what would be the solution? copy the entire thing at https://github.com/KnpLabs/KnpMenu/blob/master/src/Knp/Menu/Resources/views/knp_menu.html.twig#L101 and try to translate safe_label or label and do |raw if we translated safe_label?

stof: yeah, this is what I did at work:

```twig
{% block label %}
    {%- set translation_domain = item.extra('translation_domain', 'messages') -%}
    {%- set label = item.label -%}
    {%- if translation_domain is not same as(false) -%}
        {%- set label = label|trans(item.extra('translation_params', {}), translation_domain, locale=space_locale(item.extra('space_designation'))) -%}
    {%- endif -%}
    {%- if options.allow_safe_labels and item.extra('safe_label', false) %}{{ label|raw }}{% else %}{{ label }}{% endif -%}
{% endblock %}
```

dbu: what is the space_locale? is that something custom from your project? i would expect that we do just use the request locale here?

stof: yeah, it is specific to my project. I edited the snippet to remove it.

WouterJ:
Updated.

Resources/views/menu.html.twig
```diff
@@ -0,0 +1,10 @@
+{% extends 'knp_menu.html.twig' %}
+
+{% block label %}
+    {%- set translation_domain = item.extra('translation_domain', 'messages') -%}
+    {%- set label = item.label -%}
+    {%- if translation_domain is not same as(false) -%}
```
dbu: i think the option to set the translation domain to false to not translate at all is not documented.

stof: indeed, it is not.

WouterJ:
Rebased and documented again, should be ready now

DependencyInjection/Configuration.php
```diff
@@ -37,7 +37,7 @@ public function getConfigTreeBuilder()
                     ->addDefaultsIfNotSet()
                     ->canBeUnset()
                     ->children()
-                        ->scalarNode('template')->defaultValue('knp_menu.html.twig')->end()
+                        ->scalarNode('template')->defaultValue('KnpMenuBundle::menu.html.twig')->end()
```
dbu: is this a risk of BC break? or did the previous default simply not work and thus had to be overwritten?

dbu: ah sorry, found it. you create a new template extending that one.

why do you extend it - could this also go into the base KnpMenu? the template does not look symfony specific to me. or am i missing something?

WouterJ: Translation is symfony specific (well, twig bridge)

dbu: ok, that makes sense

Issue #236, opened 2014-07-04
Adding documentation about voters


Nek-:
- DI tags
- priority (see #281)

WouterJ:
This is now documented in the library, I don't think we have to document it in the bundle?

dbu:
i think the bundle documentation should mention and link it in some place (maybe list the features of the library in the bundle introduction?). and do we have DI tags for the voters? if so we should document them here.

stof:
We have DI tags

xDaizu:
@dbu +1

They're not documented, though. And I'm not seting a foot inside the */vendor/* folder to dig it out myself! Nothing good ever comes from meddling there!

PS: @wouterJ how to do configure stuff on a library doesn't always match the "symfony way". There's often a better way specific to the bundle. That should be documented here, in the bundle. It's not that you can't do it the other way, but it is usually better(more mantainable/painless) to stick to what the developers have in mind, so it won't break in the next minor version or edge cases or worse,you inadvertedly try to "reinvent the wheel's menu" when there's already some *hidden magic* that could do the job for you.

dbu:
@xDaizu i am not sure if anybody currently maintaining this bundle is the person who wrote that part of the code. we all dig into things and figure out stuff. open source is not only about consuming code for free, but also about everybody helping and contributing back. i am sure good things come from looking into /vendor/ and understand what your project is based on, at the least you will learn a lot. don't be afraid of breaking stuff, you will get feedback on pull requests.

xDaizu:
@dbu I know, I know, it was mainly a joke. It should be documented, though. It will make the life of newcomers like me way easier. But I intend to keep poking (both the code and the community). If people more savvy than me doesn't document it, be it a matter of time or priorities, rest assured, I'll eventually do it myself. :)

PR #281, opened 2015-09-20
Add a priority to allow controlling the order of voters

stof:
The order of voters can be important in 2 cases:

- the first voter which does not abstain wins the decision. In case multiple voters can take opposite decisions on the same item, the order is important (this will never happen for matchers shipped in the library as they either match or abstain, but never reject a match)
- the order of matchers can affect performance: if one of the matchers performs an heavy computation, running it after other matchers can avoid calling it when another matcher is taking a decision.

Replaces #268

### DIFF
--- a/DependencyInjection/Compiler/AddVotersPass.php
+++ b/DependencyInjection/Compiler/AddVotersPass.php
@@ -22,14 +22,28 @@ class AddVotersPass implements CompilerPassInterface
         $definition = $container->getDefinition('knp_menu.matcher');
         $listener = $container->getDefinition('knp_menu.listener.voters');
 
-        foreach ($container->findTaggedServiceIds('knp_menu.voter') as $id => $tags) {
-            $definition->addMethodCall('addVoter', array(new Reference($id)));
+        $voters = array();
 
+        foreach ($container->findTaggedServiceIds('knp_menu.voter') as $id => $tags) {
             foreach ($tags as $tag) {
+                $priority = isset($tag['priority']) ? (int) $tag['priority'] : 0;
+                $voters[$priority][] = $id;
+
                 if (isset($tag['request']) && $tag['request']) {
                     $listener->addMethodCall('addVoter', array(new Reference($id)));
                 }
             }
+        }
+
+        if (empty($voters)) {
+            return;
+        }
+
+        krsort($voters);
+        $sortedVoters = call_user_func_array('array_merge', $voters);
+
+        foreach ($sortedVoters as $id) {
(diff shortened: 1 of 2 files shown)


DependencyInjection/Compiler/AddVotersPass.php
```diff
@@ -22,14 +22,28 @@ public function process(ContainerBuilder $container)
...
+
+        if (empty($voters)) {
+            return;
+        }
+
+        krsort($voters);
+        $sortedVoters = call_user_func_array('array_merge', $voters);
+
+        foreach ($sortedVoters as $id) {
+            $definition->addMethodCall('addVoter', array(new Reference($id)));
```
dbu: in the cmf chain router for example we put the priority handling into the chain router. this allows to add more voters at runtime, at the cost of having to sort them at runtime. see https://github.com/symfony-cmf/Routing/blob/master/ChainRouter.php#L80-L93 and https://github.com/symfony-cmf/Routing/blob/master/ChainRouter.php#L124-L134

both have pro and con. is it a realistic scenario that i want to add my voter in a separate compiler pass or at runtime?

dbu: ah ok, just read the discussion in https://github.com/KnpLabs/KnpMenu/pull/197 so we keep as is.

stof: well, this is what I rejected in #268. I don't want to add the concept of priorities in the library itself.

If you want to add voters in your own compiler pass by modifying the definition directly, you can still manipulate the whole list of method calls of the Definition if you don't want to add it at the end of the list (same than what you would do today).

Do you really need to register a voter at runtime ? I never faced such scenario.

dbu:
this is missing documentation. please add a short note where we explain how to tag voter services.

stof:
@dbu we don't have any doc about registering voters currently. I will open a separate issue about that, but documenting the registration of voters should be its own PR.

stof:
Actually, we already have such issue: https://github.com/KnpLabs/KnpMenuBundle/issues/236

dbu:
agreed. added a note in #236

travis seems rather confused with the push build. either its talking to the wrong repo or there is a caching issue. i tried restarting the build but same outcome. the pr build is green however. is that enough? imho we can merge this.

stof:
@dbu I pushed to the wrong remote first and delete the branch and pushed back to my fork. but Travis triggered a build for the push to the main repo, and it was the same commit.

PR #285, opened 2015-10-01
Fix documentation about MenuBuilder as a service

ugomeda:
Options sent to the MenuBuilder are in 3rd position when calling knp_menu_get


stof:
thanks

PR #286, opened 2015-10-19
Passing second argument to createItem method.

BartoszBartniczak:


stof:
-1 for that. Menu builder options are not meant to be the options of the root item directly, and I don't want to advocate this pattern.
the menu builder options are meant to be used the way your menu builder wants

BartoszBartniczak:
OK. I understand your point. Thank you for your help.

WouterJ:
Closing as we agreed this was a not-so-good practice.

PR #289, opened 2015-11-09
Allow Symfony 3

WouterJ:


stof:
Tests need to be fixed to stop using the ContainerAware class (you can implement the interface manually to avoid it, as it is removed in 3.0. It is a fixture class anyway)

WouterJ:
@stof well, the problem is the [`ContainerAwareProvider`](https://github.com/KnpLabs/KnpMenuBundle/blob/master/Renderer/ContainerAwareProvider.php) provided by the bundle, which extends `ContainerAware`.

Is it maybe possible to change that extend to only implement `ContainerAwareInterface` and add the missing method manually? Or is it considered a BC break? (typehinting ContainerAware instead of ContainerAwareInterface seems a bit strange)

stof:
ContainerAwareProvider does not extends ContainerAware. The issue is in ``Knp\Bundle\MenuBundle\Tests\Stubs\Menu\ContainerAwareBuilder``, which is clearly not covered by BC as it is a test stub.

Btw, there are a few places in the doc which are using ContainerAware too, which should be updated to talk about implementing the interface instead. See https://github.com/KnpLabs/KnpMenuBundle/issues/290 which has a link to the search for them

WouterJ:
Builds are passing now. Thanks @stof for the guidance!

stof:
Thank you @WouterJ.

Tests/Stubs/Menu/ContainerAwareBuilder.php
```diff
@@ -3,10 +3,18 @@
 namespace Knp\Bundle\MenuBundle\Tests\Stubs\Menu;
 
 use Knp\Menu\FactoryInterface;
-use Symfony\Component\DependencyInjection\ContainerAware;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
 
-class ContainerAwareBuilder extends ContainerAware
+class ContainerAwareBuilder implements ContainerAwareInterface
```
BallisticPain: From my understanding of the changes coming this does not get it ready for Symfony 3.0. ContainerAwareInterface is what is deprecated. You need to `use ContainerAwareTrait;` is the recommended fix.

Thoughts? @WouterJ & @stof

PR #298, opened 2015-12-15
Update documentation to Symfony 3 compatibility

JHGitty:


Resources/doc/index.rst
```diff
@@ -177,12 +188,6 @@ With the standard ``knp_menu.html.twig`` template and your current page being
 
 .. note::
 
-    You only need to extend ``ContainerAware`` if you need the service
-    container to be available via ``$this->container``. You can also implement
```
dbu: i think the first part of this note is that you don't need to do ContainerAwareInterface unless you actually need the container. this is still true and we should keep that part of the note.

JHGitty:
Currently I don't have much time to think about this issue. Anybody could fix it.

dbu:
did #299

Issue #290, opened 2015-11-12
Use ContainerAwareTrait instead of ContainerAware


JHGitty:
Found usages of `ContainerAware`:
https://github.com/KnpLabs/KnpMenuBundle/search?utf8=%E2%9C%93&q=ContainerAware

Outdated documentation:
http://symfony.com/doc/master/bundles/KnpMenuBundle/index.html#method-a-the-easy-way-yay

Deprecation:
The AdminBundle\Menu\Builder class extends     Symfony\Component\DependencyInjection\ContainerAware that is deprecated since version 2.8, to be removed in 3.0. Use the ContainerAwareTrait instead.

JHGitty:
Fixed by #299

PR #299, opened 2015-12-17
adjust documentation to be symfony 3 compatible

dbu:
symfony 3 does not have a ContainerAware to extend. but anyway that is bad practice and not needed in this case.

continue from #298

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -125,10 +125,13 @@ An example builder class would look like this:
     namespace AppBundle\Menu;
 
     use Knp\Menu\FactoryInterface;
-    use Symfony\Component\DependencyInjection\ContainerAware;
+    use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+    use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 
-    class Builder extends ContainerAware
+    class Builder implements ContainerAwareInterface
     {
+        use ContainerAwareTrait;
+        
         public function mainMenu(FactoryInterface $factory, array $options)
         {
             $menu = $factory->createItem('root');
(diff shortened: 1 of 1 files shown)


Resources/doc/index.rst
```diff
@@ -124,11 +124,18 @@ An example builder class would look like this:
...
+    use Doctrine\Common\Persistence\ManagerRegistry;
     use Knp\Menu\FactoryInterface;
-    use Symfony\Component\DependencyInjection\ContainerAware;
 
-    class Builder extends ContainerAware
+    class Builder
     {
+        private $registry;
+        
+        public function __construct(ManagerRegistry $registry)
```
stof: This is wrong. This example is about documenting menus rendered as ``knp_menu_render('AppBundle:Builder:mainMenu')``, so you cannot inject the ManagerRegistry in the constructor. This should be using ContainerAwareInterface here.

Registering builders as services is documented in a different chapter

JHGitty:
@stof Sorry, deleted my text already - your answer was too fast. Thanks! :+1:

dbu:
oh, right you are stof. updated it. better?

BallisticPain:
Gents,

I thought the ContainerAwareInterface was being deprecated as well? Maybe I'm mistaken.

I was just looking to work on this based on the Issue #290 
So I believe this fixes #290 pending my above question.

Thoughts?

JHGitty:
@BallisticPain The ``ContainerAwareInterface`` is **not** deprecated because it is useful in combination with ``ContainerAwareTrait``.

stof:
@JHGitty actually more the opposite: ContainerAwareTrait is useful when you need to implement ContainerAwareInterface :smile:

JHGitty:
@stof Yes, thanks. :-)

PR #308, opened 2016-02-22
Fixed a code block directive

javiereguiluz:
`xliff` is not a valid value for the `code-block` directive. Let's use `xml` instead: 

Error details:

```
Exception occurred:
File "sphinx/configurationblock.py", line 60, in run
  innernode = nodes.emphasis(self.formats[child['language']], self.formats[child['language']])
  KeyError: u'xliff'
```


WouterJ:
Thanks!

PR #311, opened 2016-03-07
Update index.rst

koku:
Without an echo statement, not much is outputted.


WouterJ:
Great catch, thanks @koku!

PR #317, opened 2016-04-22
handle alias for factory and add sanity checks for it

dbu:
in https://github.com/symfony-cmf/menu-bundle/issues/249 we noticed that when we replace knp_menu.factory with an alias, the compiler pass stops working. once i fixed that, i noticed that this compiler pass assumes a method that is not in the FactoryInterface.

with these changes, we can handle an alias and we fail with a better error message than a fatal error at run time when the factory is instantiated.

### DIFF
--- a/DependencyInjection/Compiler/AddExtensionsPass.php
+++ b/DependencyInjection/Compiler/AddExtensionsPass.php
@@ -2,6 +2,7 @@
 
 namespace Knp\Bundle\MenuBundle\DependencyInjection\Compiler;
 
+use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\Reference;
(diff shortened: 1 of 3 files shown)


dbu:
either we now catch something with the test setup, or i missed something that the tests reveal.

i will wait for feedback if the approach itself is viable before i dig further into this.

DependencyInjection/Compiler/AddExtensionsPass.php
```diff
@@ -15,13 +16,19 @@ class AddExtensionsPass implements CompilerPassInterface
...
-        if (!$container->hasDefinition('knp_menu.factory')) {
+        if (!$container->has('knp_menu.factory')) {
             return;
         }
 
-        $definition = $container->getDefinition('knp_menu.factory');
+        $definition = $container->findDefinition('knp_menu.factory');
 
         foreach ($container->findTaggedServiceIds('knp_menu.factory_extension') as $id => $tags) {
+            if (!method_exists($definition->getClass(), 'addExtension')) {
```
stof: you need to resolve the value for the class, as it may use parameters.

stof: and this check should be done only once, not for each extension

DependencyInjection/Compiler/AddExtensionsPass.php
```diff
@@ -15,13 +16,25 @@ class AddExtensionsPass implements CompilerPassInterface
...
 
-        $definition = $container->getDefinition('knp_menu.factory');
+        $taggedServiceIds = $container->findTaggedServiceIds('knp_menu.factory_extension');
+        if (0 === count($taggedServiceIds)) {
+            return;
+        }
+
+        $definition = $container->findDefinition('knp_menu.factory');
+
+        if (!method_exists($container->resolveServices($definition)->getClass(), 'addExtension')) {
```
dbu: is this the right way to resolve?

dbu:
thanks stof. updated.

efeencheung:
This is better than a fatal error，But if some one did not make a right bundle order，they also get a exception，which confused them.

dbu:
we can't do better on this end of the code. silently ignoring the
problem is much worse than an exception. and we only throw the exception
if there actually is a tagged service.

the rest of the fix is in the cmf menu bundle where we implement the
addExtension method. once both are merged, the order of bundle
instantiation is no longer relevant.

efeencheung:
It works now，Before next release，I can change the bundle order to continue my work，thanks @dbu

DependencyInjection/Compiler/AddExtensionsPass.php
WouterJ: I would do `$container->getParameterBag()->resolveValue($definition->getClass())`

dbu:
thanks @WouterJ , only resolving the class now instead of the whole definition.

WouterJ:
@dbu can you please update the `Knp\Bundle\MenuBundle\Tests\DependencyInjection\Compiler\AddExtensionsPassTest` test?

dbu:
updated the test and did some cleanup discovered while doing the test.

WouterJ:
:+1:

dbu:
ping - can we merge this (and https://github.com/symfony-cmf/menu-bundle/pull/251?)

dbu:
@stof do you see any issue with this or is it ok to merge?

PR #318, opened 2016-04-25
Fix wrong event name in Resources/doc/events.rst

martinsik:
This is probably a leftover from https://github.com/KnpLabs/KnpMenuBundle/commit/8cf6329e63e4bb8a14336c6facaa8d5f48f80f52.


WouterJ:
Thanks @martinsik, nice catch! (good detective work too btw).

PR #319, opened 2016-04-26
Removed unused dependency from events.rst

martinsik:
I've noticed this class is neither defined nor used anywhere, it looks like it was predecessor of `AppBundle\Event\ConfigureMenuEvent`.

### DIFF
--- a/Resources/doc/events.rst
+++ b/Resources/doc/events.rst
@@ -17,7 +17,6 @@ to allow other parts of your application to add more stuff to it.
 
   namespace AppBundle\Menu;
 
-  use AppBundle\MenuEvents;
   use AppBundle\Event\ConfigureMenuEvent;
   use Knp\Menu\FactoryInterface;
   use Symfony\Component\DependencyInjection\ContainerAware;


martinsik:
In the same file, there's also some `Symfony\Component\DependencyInjection\ContainerAware` but this class doesn't exist. Was this supposed to be `Symfony\Component\DependencyInjection\ContainerAwareInterface` or `Knp\Bundle\MenuBundle\Tests\Stubs\Menu\ContainerAwareBuilder`? Inside `ContainerAwareBuilder` there's `container` property but it's `private` so I'm confused.

stof:
@martinsik ContainerAware exists in Symfony 2.x. It has been removed in Symfony 3 (in favor of implementing the interface directly, optionally using the provided trait). This doc is older than the deprecation of the class

dbu:
@stof but is this fix in the PR still correct? should we change the container aware mention to the interface and trait?

stof:
yeah, we should probably update the doc to implement the interface rather than extending the abstract class (as this is the actual need).

removing this unused use statement is indeed correct, so I'm merging this PR.

PR #332, opened 2016-11-24
Replace ContainerAware with ContainerAwareInterface

bigstylee:
The class Symfony\Component\DependencyInjection\ContainerAware does not exist.


stof:
Thanks

PR #333, opened 2016-11-27
"Removed the methods related to the current uri in ItemInterface "

bigstylee:
https://github.com/KnpLabs/KnpMenu/commit/e7ef14a9469545423276e8516196496664429b81


stof:
Good catch, thanks

PR #337, opened 2017-02-18
Complex Translation Structure Examples

BallisticPain:
I ran into a brick wall, and this documentation didn't help. I looked into the Forms component, and thought it should work the same, tested it, and here I am sending a PR! Woot!

**NOTE** I don't have the knowledge to create my _complex_ structure in the xlf format. If someone could please inform me, or add a PR here that would be awesome!

### DIFF
--- a/Resources/doc/i18n.rst
+++ b/Resources/doc/i18n.rst
@@ -47,6 +47,50 @@ The items "Home" and "Login" can now be translated in the message domain:
             'Login' => 'Connexion',
         );
 
+More Complex Translation Structure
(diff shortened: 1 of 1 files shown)


stof:
XML translation files are using the Xliff format, so there is no nested syntax here. The XML file has to use the actual keys (so `dashboard.tabs.home`) in this case.

However, I'm against adding this doc in the bundle doc. KnpMenuBundle is not responsible for documenting the Symfony translation system. This is already done in the official Symfony documentation.

PR #341, opened 2017-03-06
Add templating component

garak:
Templating component is no longer required by symfony/framework-bundle, so we need to require it

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,6 +27,9 @@
         "symfony/expression-language": "~2.4|~3.0",
         "symfony/templating": "~2.3|~3.0"
     },
+    "suggest": {
+        "symfony/templating": "When not using Twig but the Symfony templating component to render menus. (~2.3|~3.0)"
+    },
     "autoload":     {
         "psr-4": { "Knp\\Bundle\\MenuBundle\\": "" }
     },
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,8 @@
     ],
     "require":      {
         "knplabs/knp-menu":         "~2.2",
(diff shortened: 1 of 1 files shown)


stof:
should be a dev requirement only actually. You don't need the templating component when using Twig

dbu:
do we need to update the documentation as well?

and this could be added into a `suggests` section to have it shown by composer that you might want this. something like

```
    "suggest": {
         ""symfony/templating": "When not using Twig but the Symfony templating component to render menues. (~2.3|~3.0)"
    }
```

garak:
Is there really someone that uses Symfony without Twig?

dbu:
i am pretty sure there are. i hope i don't meet them (unless there is a very good and very interesting reason) but the bundle should not assume.

i would have no issue deprecating the templating component support and drop it in the next major version. (if somebody actually needs it, the knp menu does have a renderer interface as well as other plain renderers, which could be used as a replacement). then we should not have the suggest, but the doc for the templating support should still mention that you need to require the component but also say that that is deprecated in favor of twig.

@stof does that make sense or do you see good reason to keep php templating support in this bundle?

stof:
I would remove the suggest: telling you to install the templating component to get an integration when you use PHP templates is totally useless: if you do use PHP templates, you *already* have the component, and so the suggestion won't be displayed by Composer.

stof:
OK, as the suggestion was added as a separate commit, I made a manual merge of only the first 2 commits.
I'm closing the PR (Github won't mark it as merged as the last commit adding the suggestion is not merged, but your contribution is preserved for the 2 other commits)

PR #346, opened 2017-06-09
Update travis.yml with PHP 7.1

DonCallisto:
As PHP 7.1 is out since December (6 months) I think is useful to test the codebase against it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ php:
     - 5.5
     - 5.6
     - 7.0
-    - hhvm
+    - 7.1
 
 matrix:
     include:
@@ -24,6 +24,8 @@ matrix:
           env: SYMFONY_VERSION='2.3.*'
         - php: 5.6
           env: SYMFONY_VERSION='3.0.*@dev'
+        - php: hhvm
+          dist: trusty
 
 before_install:
     - composer self-update # remove this once the Travis upgrade of February 2015 is deployed


DonCallisto:
There are some issues

One related to ubuntu and HHVM

> HHVM is no longer supported on Ubuntu Precise. Please consider using Trusty with `dist: trusty`.

I need to figure out others.

Let me know how should I move on (for HHVM), if you're interested in 7.1. For others, I'll take a look ASAP if you confirm your interest in this PR

dbu:
yes. definitely makes sense to test on php 7.1. can you please do the same on https://github.com/KnpLabs/KnpMenu ?

i think the other problem we are seeing is that the pre-installed phpunit bumped to the new major version which uses namespaces... the approach i have seen most is adding phpunit to require-dev and run vendor/bin/phpunit rather than whatever happens to be installed on the build system.

DonCallisto:
> can you please do the same on https://github.com/KnpLabs/KnpMenu ?

Of course, I'll do in next hours

> i think the other problem we are seeing is that the pre-installed phpunit bumped to the new major version which uses namespaces

That's right

> he approach i have seen most is adding phpunit to require-dev and run vendor/bin/phpunit rather than whatever happens to be installed on the build system.

Which version? Last one? Because I suppose in that case when need to change tests.
BTW, this should be a separate PR, right? Who will go? Would you like me to do that fix?

PR #348, opened 2017-08-22
Improve the CI setup

stof:
- use the simple-phpunit script instead of the Travis-provided phar to control the PHPUnit version
- ensure PHP 5.3 job is ready for the Travis migration to Trusty
- Clean the testing of Symfony version (no more testing of 2.8 for the dev job, and keep only LTS as explicit jobs)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,14 @@ sudo: false
 cache:
     directory:
         - $HOME/.composer/cache/files
+        - $HOME/symfony-bridge/.phpunit
+
+env:
+    global:
+        - SYMFONY_PHPUNIT_DIR="$HOME/symfony-bridge/.phpunit"
+        - SYMFONY_PHPUNIT_REMOVE="symfony/yaml" # keep Prophecy
 
 php:
-    - 5.3
     - 5.4
     - 5.5
     - 5.6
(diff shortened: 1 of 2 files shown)


stof:
this should make the CI green again

.travis.yml
```diff
@@ -5,9 +5,14 @@ sudo: false
 cache:
     directory:
         - $HOME/.composer/cache/files
+        - $HOME/symfony-bridge/.phpunit
+
+env:
+    global:
+        - SYMFONY_PHPUNIT_DIR="$HOME/symfony-bridge/.phpunit"
```
stof: defining this env variable allows to make the bridge install the PHPUnit in a cacheable location instead of downloading it in `vendor/symfony/phpunit-bridge/.phpunit` for which Travis caching of the folder would interfer with the Composer installation.

Btw, for local development, I added this in my ZSH config:

```bash
# Symfony PHPUnit bridge
export SYMFONY_PHPUNIT_DIR=~/.cache/symfony-phpunit-bridge
export SYMFONY_PHPUNIT_REMOVE=symfony/yaml
```

This is a nice way to share the PHPUnit downloaded by the bridge among all projects I have locally

PR #351, opened 2017-08-22
Add validation of the composer.json on Travis

stof:
This prevents PRs breaking the validity of the composer.json, which would then be rejected by Packagist.
The lock file is not validated, as it is not committed in the repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,4 +41,6 @@ install:
     - composer update $COMPOSER_FLAGS
     - ./vendor/bin/simple-phpunit install
 
-script: ./vendor/bin/simple-phpunit -v
+script:
+    - composer validate --strict --no-check-lock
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -41,4 +41,6 @@ install:
     - composer update $COMPOSER_FLAGS
     - ./vendor/bin/simple-phpunit install
 
-script: ./vendor/bin/simple-phpunit -v
+script:
+    - composer validate --strict --no-check-lock
```
dbu: doesn't `composer update` also validate the composer.json file?

stof: well, it breaks if the file is totally invalid, which would break the update.
But it does not break if the file has warnings making it unsuitable for publication on Packagist. And this is what happened in the past in some repos I maintain (stopping updates of the master branch in Packagist).

PR #353, opened 2017-08-22
Add support for autoconfiguring menu voters

stof:
This allows people to skip the tag `knp_menu.voter` for their VoterInterface implementations if their services are marked as autoconfigured.

### DIFF
--- a/DependencyInjection/Compiler/AddVotersPass.php
+++ b/DependencyInjection/Compiler/AddVotersPass.php
@@ -25,6 +25,10 @@ class AddVotersPass implements CompilerPassInterface
         $voters = array();
 
         foreach ($container->findTaggedServiceIds('knp_menu.voter') as $id => $tags) {
+            // Process only the first tag. Registering the same voter multiple time
+            // does not make any sense, and this allows user to overwrite the tag added
+            // by the autoconfiguration to change the priority (autoconfigured tags are
+            // always added at the end of the list).
             $tag = $tags[0];
 
             $priority = isset($tag['priority']) ? (int) $tag['priority'] : 0;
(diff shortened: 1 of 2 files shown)


DependencyInjection/Compiler/AddVotersPass.php
```diff
@@ -25,13 +25,13 @@ public function process(ContainerBuilder $container)
         $voters = array();
 
         foreach ($container->findTaggedServiceIds('knp_menu.voter') as $id => $tags) {
-            foreach ($tags as $tag) {
-                $priority = isset($tag['priority']) ? (int) $tag['priority'] : 0;
-                $voters[$priority][] = $id;
+            $tag = $tags[0];
```
dbu: are you sure about this? if its possible to add more than one of the same tag, people might have done this. should we raise an error if `count($tags) > 1`?

stof: well, for voters, there is no point registering the same voter multiple times (except for hurting performance, but I don't consider this a supported feature, and I doubt anyone wants such feature). The second vote will be the same than the first vote.

And we don't want to forbid having multiple tags. We want to ignore subsequent tags. This is what allows you to override the autoconfigured tag without opting out of all other autoconfiguration: you can add your own tag with a priority, and it will win over the tag added by the autoconfiguration at a priority of 0.
This is consistent with what Symfony does btw. Any tag used to collect services that don't make sense to be collected multiple times is now processing only the first tag. Tags which are used to attach additional info (these often have mandatory attributes) are processing all tags and are not autoconfigured (they cannot be anyway if some attributes are mandatory)

dbu: ah, cool! that makes a lot of sense then. i would add a short comment in this code with that explanation so that devs looking at this understand its relevant that we only process the first tag.

PR #358, opened 2017-10-02
Update configuration for KnpMenu 2.3 and Symfony 3.3+

stof:
This PR uses all the new features I added recently to benefit from Symfony 3.3 DI features (allowing to keep menu builders and menus as private services for instance).

Support for older Symfony versions is kept (I want to have at least 1 release of the bundle compatible with the 2.x LTS and supporting KnpMenu 2.3 without deprecations), but support for private services in these places is not available in this case.

Each commit performs a different refactoring, so things are easier to review commit by commit (I could have split these into multiple PRs, but it was harder for me as I was in the train)

### DIFF
--- a/DependencyInjection/Compiler/AddProvidersPass.php
+++ b/DependencyInjection/Compiler/AddProvidersPass.php
@@ -1,6 +1,7 @@
 <?php
 namespace Knp\Bundle\MenuBundle\DependencyInjection\Compiler;
 
+use Symfony\Component\DependencyInjection\Argument\IteratorArgument;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\Reference;
@@ -30,6 +31,10 @@ class AddProvidersPass implements CompilerPassInterface
             // when using only one (the default case as the bundle defines one provider)
             $container->setAlias('knp_menu.menu_provider', (string) reset($providers));
         } else {
+            if (class_exists(IteratorArgument::class)) {
+                $providers = new IteratorArgument($providers);
+            }
+
             $definition = $container->getDefinition('knp_menu.menu_provider.chain');
             $definition->replaceArgument(0, $providers);
             $container->setAlias('knp_menu.menu_provider', 'knp_menu.menu_provider.chain');
--- a/DependencyInjection/Compiler/AddProvidersPass.php
+++ b/DependencyInjection/Compiler/AddProvidersPass.php
@@ -9,6 +9,8 @@ use Symfony\Component\DependencyInjection\Reference;
  * This compiler pass registers the providers in the ChainProvider.
  *
  * @author Christophe Coevoet <stof@notk.org>
+ *
+ * @internal
  */
 class AddProvidersPass implements CompilerPassInterface
 {
--- a/Renderer/ContainerAwareProvider.php
+++ b/Renderer/ContainerAwareProvider.php
@@ -5,17 +5,24 @@ namespace Knp\Bundle\MenuBundle\Renderer;
 use Knp\Menu\Renderer\RendererProviderInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
+/**
+ * @deprecated since version 2.2. Use "Knp\Menu\Renderer\PsrProvider" instead.
+ */
 class ContainerAwareProvider implements RendererProviderInterface
 {
     private $container;
     private $rendererIds;
     private $defaultRenderer;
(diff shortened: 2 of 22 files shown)


dbu:
:+1: 

as noted in #359, symfony 3.3 support also means we should require twig in composer.json or gracefully handle the absence of twig. symfony/framework-bundle stopped depending on twig...

DependencyInjection/Compiler/AddProvidersPass.php
```diff
@@ -28,6 +31,10 @@ public function process(ContainerBuilder $container)
             // when using only one (the default case as the bundle defines one provider)
             $container->setAlias('knp_menu.menu_provider', (string) reset($providers));
         } else {
+            if (class_exists(IteratorArgument::class)) {
```
dbu: is this symfony 3.3+ only? want to add a comment so we know we can drop the exists check when we drop support for older versions of symfony?

Renderer/ContainerAwareProvider.php
```diff
@@ -5,17 +5,24 @@
...
 
-    public function __construct(ContainerInterface $container, $defaultRenderer, array $rendererIds)
+    public function __construct(ContainerInterface $container, $defaultRenderer, array $rendererIds, $triggerDeprecation = true)
     {
         $this->container = $container;
         $this->rendererIds = $rendererIds;
         $this->defaultRenderer = $defaultRenderer;
+
+        if ($triggerDeprecation) {
+            @trigger_error(sprintf('The %s class is deprecated since 2.2 and will be removed in 3.0. USe "Knp\Menu\Renderer\PsrProvider" instead.', __CLASS__),E_USER_DEPRECATED);
```
dbu: s/USe/Use/

Tests/DependencyInjection/Compiler/AddVotersPassTest.php
```diff
@@ -27,15 +28,63 @@ public function testProcessWithAlias()
...
+
+        if (class_exists(IteratorArgument::class)) {
+            $voters = new IteratorArgument($voters);
+        }
+
+        $definitionMock->expects($this->once())
+            ->method('replaceArgument')
+            ->with(0, $voters);
+
+        $listenerMock = $this->getMockBuilder('Symfony\Component\DependencyInjection\Definition')
```
dbu: i would use the ::class pseudo constant for these. and can't you do `createMock` instead of getMockBuilder->disableOriginalConstructor()->getMock()? with the phpunit bridge we should have a recent enough version of phpunit i think.

DependencyInjection/Compiler/AddProvidersPass.php
stof: yep, the class was added in 3.3

Tests/DependencyInjection/Compiler/AddVotersPassTest.php
stof: Well, the argument for using a recent version of PHPUnit is our drop of PHP 5.5 actually. Even when using the bridge, you would get PHPUnit 4.8 when running on PHP 5.5 (as PHPUnit 5 requires PHP 5.6+)

DependencyInjection/Compiler/AddProvidersPass.php
stof: And this is part of the WIP commit. I'm not done yet on these.

piotrantosik:
hi @stof, any progress on this pr?

ElectricMaxxx:
@stof could you pleas make `knp_menu.renderer.twig` public too?  As the `ContainerAwareProvider`  fetches renderer by using the container and this services is fetched that way, i got an issue in `symfony-cmf/menu-bundle`: https://travis-ci.org/symfony-cmf/menu-bundle/jobs/302709508 while trying to lift it to symfony 4.0

ElectricMaxxx:
Anything we can help here?

stof:
@ElectricMaxxx ContainerAwareProvider is not used on Symfony 3.3+ in this PR. This PR is precisely about using the new DI features *instead of* making everything public.
I'm working on this PR now during the SymfonyCon Hackday, to finish it.

ElectricMaxxx:
Serms there is one deprecation left `BundleInterface::getParent()`

stof:
@ElectricMaxxx I'm actually considering deprecating BuilderAliasProvider itself (using menu builders is easier IMO, as it allows to use autowiring while your class using the BuilderAliasProvider can only access public services from the container and most of them are not public anymore).
I even have one feature in my pocket (coming soon, I had no time to work on it during the hackday) to make it dead easy to autoconfigure menu builders (think event subscriber vs event listener for the idea)

PR #360, opened 2017-10-24
SF4 support

Zausenec:
The "knp_menu.renderer.twig" service or alias has been removed or inlined when the container was compiled. You should either make it public, or stop using the container directly and use dependency injection instead.


stof:
-1 for that. https://github.com/KnpLabs/KnpMenuBundle/pull/358 is refactoring the bundle to avoid the need to have it as a public service. I will finish it next week.

PR #362, opened 2017-11-16
add Builder Alias symfony3.3 support

Kronhyx:


stof:
What was the issue with Symfony 3.3 ?

Kronhyx:
when I call knp_menu_get ('App:CustomMenu:Locale') he does not recognize that App because it is not a bundle and an exception is generated

stof:
Well, what you are asking for is not Symfony 3.3+ support. It is support for bundle-less apps. This is indeed not supported yet. But I'm not sure I want to support this either. Registering menu builders as services is much better IMO (and for Symfony 4, I will soon provide a way to have them taking profit of autoconfiguration on Symfony 3.3+)

Kronhyx:
Well, what am I supposed to do? Many developers use applications without packages and probably present this problem in the future

Kronhyx:
ok @stof , thx for all

Issue #364, opened 2017-11-20
$voter->setRequest() deprecated in KnpMenu


mgiraud:
Hello,

The `VoterInitializerListener` sets the request to voters : https://github.com/KnpLabs/KnpMenuBundle/blob/master/EventListener/VoterInitializerListener.php#L16
```php
public function onKernelRequest(GetResponseEvent $event)
    {
        if (HttpKernelInterface::MASTER_REQUEST !== $event->getRequestType()) {
            return;
        }
        foreach ($this->voters as $voter) {
            if (method_exists($voter, 'setRequest')) {
                $voter->setRequest($event->getRequest());
            }
        }
    }
```

However the `setRequest` is deprecated in the KnpMenu 2.3 : https://github.com/KnpLabs/KnpMenu/blob/master/src/Knp/Menu/Matcher/Voter/RouteVoter.php#L51

which triggers deprecation warnings when using phpunit

stof:
Fixed by #358

PR #365, opened 2017-11-21
Resolve KnpMenu 2.3 deprecations

mgiraud:
This is a quick PR to resolve the KnpMenu 2.3 deprecations as described in #364 . Maybe needs more work and contributions. 

- Remove the VoterInitializerListener.php
- Use service definition in Compiler Pass to inject the request_stack service
- Fix composer (missing Prophecy), and requires knp-menu:~2.3


composer.json
```diff
@@ -20,13 +20,14 @@
...
-        "knplabs/knp-menu":         "~2.2",
+        "knplabs/knp-menu":         "~2.3",
         "symfony/framework-bundle": "~2.7|~3.0"
     },
     "require-dev":  {
         "symfony/phpunit-bridge": "^3.3",
         "symfony/expression-language": "~2.7|~3.0",
-        "symfony/templating": "~2.7|~3.0"
+        "symfony/templating": "~2.7|~3.0",
+        "phpspec/prophecy": "^1.7"
```
stof: PHPUnit includes Prophecy in the phar since 4 years now.

stof:
Closing this in favor of #358, which does much more than that, and preserve backward compatibility (your PR is a BC break because custom voters may use the tag as well).

stof:
@Nek- the fact that the PR on the bundle was not merged yet was precisely the reason why I was waiting before releasing the library, to prevent such PR updating the bundle in parallel. Too bad that you released the library a few days ago :(

Issue #374, opened 2017-12-13
Refactor to work with twig directly when the templating component is not used


dbu:
as said in #372: This bundle needs a bugfix to properly register the path in case Twig is used without using the Templating component.

codedmonkey:
Is there a particular reason this bundle uses `twig.loader.filesystem` instead of `twig.loader.native_filesystem`? It seems that could easily circumvent the Templating dependency entirely since that one is included in the Twig bundle. During my short debugging it worked fine.

PR #377, opened 2017-12-19
Refactor hard dependency on the Templating component with a fallback …

codedmonkey:
Refactors the hard dependency on the Templating component with a fallback option on the TwigBundle as discussed in https://github.com/KnpLabs/KnpMenuBundle/issues/374.


stof:
Closing in favor of #378 which fixes it in a different way.

PR #381, opened 2018-01-22
Update .travis.yml

andreybolonin:


.travis.yml
```diff
@@ -30,6 +30,10 @@ matrix:
           env: SYMFONY_VERSION='^2'
         - php: 7.1
           env: SYMFONY_VERSION='^3'
+        - php 7.1
```
dbu: you need `- php: 7.1` (missing colon)
same with the next entry

andreybolonin: ok, one moments, thx

andreybolonin: tests passed

.travis.yml
```diff
@@ -30,10 +30,13 @@ matrix:
...
         - php: 7.1
           env: SYMFONY_VERSION='^3'
+        - php: 7.1
+          env: SYMFONY_VERSION="4.0.*"
+        - php: 7.2
+          env: SYMFONY_VERSION="4.0.*"
 
 before_install:
     - if [ "$DEPENDENCIES" == "dev" ]; then composer config minimum-stability dev; fi;
-    - if [ "$SYMFONY_VERSION" != "" ]; then composer require symfony/lts:${SYMFONY_VERSION} --no-update; fi;
```
dbu: this makes us ignore the SYMFONY_VERSION, we need to keep this line in

.travis.yml
```diff
@@ -30,10 +30,13 @@ matrix:
           env: SYMFONY_VERSION='^2'
         - php: 7.1
           env: SYMFONY_VERSION='^3'
+        - php: 7.1
+          env: SYMFONY_VERSION="4.0.*"
+        - php: 7.2
+          env: SYMFONY_VERSION="4.0.*"
```
stof: -1 for that. There is no reason to force testing the 4.0.x minor version explicitly. Currently, it is tested by normal jobs as being the latest stable version. And at the time these jobs are switching to Symfony 4.1, Symfony 4.0 will reach end of maintenance 2 months later, so not worth having us update this config every 6 months (and wait longer on Travis due to the quota on concurrent jobs per github organization).

The comment before these jobs says `Test LTS version`. 4.0 is not a LTS.

and btw, testing SF 4 on multiple PHP version version is even less relevant regarding our Travis quota

stof:
I'm closing this PR. See my comment above about why it does not make sense.

PR #385, opened 2018-03-06
Add autowiring for MenuProviderInterface

### LINKED ISSUE #384: Provide autowiring for MenuProviderInterface
Current error:

```
Cannot autowire service "AppBundle\Menu\SidebarMenuBuilder": argument "$menuProvider" of method "__construct()" references interface "Knp\Menu\Provider\MenuProviderInterface" but no such service exists. You should maybe alias this interface to one of these existing services: "knp_menu.menu_provider.chain", "knp_menu.menu_provider.lazy", "knp_menu.menu_provider.builder_alias", "sonata.admin.menu.group_provider". Did you create a class that implements this interface?
```

Current workaround:

```yaml
services:
    _defaults:
        autowire: true
        autoconfigure: true
        public: false

    Knp\Menu\Provider\MenuProviderInterface: '@knp_menu.menu_provider'
```

Soullivaneuh:
Closes #384

### DIFF
--- a/src/Resources/config/menu.xml
+++ b/src/Resources/config/menu.xml
@@ -38,6 +38,8 @@
             <argument type="collection" />
         </service>
 
+        <service id="Knp\Menu\Provider\MenuProviderInterface" alias="knp_menu.menu_provider" public="false"/>
+
         <service id="knp_menu.menu_provider.chain" class="%knp_menu.menu_provider.chain.class%" public="false">
             <argument type="collection" />
         </service>


src/DependencyInjection/Compiler/AddProvidersPass.php
```diff
@@ -39,5 +40,6 @@ public function process(ContainerBuilder $container)
             $definition->replaceArgument(0, $providers);
             $container->setAlias('knp_menu.menu_provider', 'knp_menu.menu_provider.chain');
         }
+        $container->setAlias(MenuProviderInterface::class, 'knp_menu.menu_provider');
```
stof: Please add this in the XML file with other autowiring aliases instead.

Soullivaneuh:
It's done. I tried to do a reverse alias and deprecate `knp_menu.menu_provider`, but you apparently can't deprecate an alias.

Issue #375, opened 2017-12-13
Quick win - Fixing 3.4 deprecated warning


stphane:
Fix [deprecation message](https://github.com/symfony/symfony/blob/020664e1efbf8a81dbcc0f7b9dcbfa9ac58c0ed7/src/Symfony/Component/HttpKernel/Kernel.php#L237) introduced in Symfony 3.4

My attempt to push a new branch to issue a new PR was denied and so here might be the tiny change required.

```
--- a/src/Provider/BuilderAliasProvider.php
+++ b/src/Provider/BuilderAliasProvider.php
@@ -101,7 +101,7 @@ class BuilderAliasProvider implements MenuProviderInterface
             $logs = array();
             $bundles = array();
 
-            $allBundles = $this->kernel->getBundle($bundleName, false);
+            $allBundles = $this->kernel->getBundle($bundleName);
 
             // In Symfony 4, bundle inheritance is gone, so there is no way to get an array anymore.

```

stof:
That's not a quick win, as it breaks support for bundle inheritance entirely (and people still using it in their project would then not be happy).

stof:
And you cannot push branches directly to this repository. Only maintainers can (as for any project on github). See https://guides.github.com/activities/forking/ for the doc about it.

stphane:
Thank you for your feedback. So, such warnings will remain in the dev bar until we update our project to symfony 4 and future coming releases of KnpMenuBundle ?

dhensen:
@stphane if you update to symfony 4 now I believe that the KnpMenuBundle will be broken because of the deprecated 'false' in the getBundle call. A better fix with BC for 2.x and 3.x needs to be made or maybe a major update of KnpMenuBundle needs to be made to only support the new way of calling getBundle and drop the support for 2.x and 3.x in this new major version of KnPMenuBundlE

stof:
no, it will not be broken with Symfony 4. The code already handles the case where an array is not returned anymore (and PHP itself does not forbid passing an extra argument)

stphane:
@dhensen I do agree with this. (My question was, in other words, «Will there be an update to this bundle for 3.4.x to drop support for previous Symfony versions ?» Warnings would then disappear…)

stof:
@stphane Symfony 3.4 also supports bundle inheritance, even though it is deprecated. So removing this code entirely would be done only when dropping support for SF 3.4 itself too.

geelweb:
[...]

PR #387, opened 2018-04-09
Fix deprecation notice on Symfony 3.4 and higher, assuring BC

ricardocorreia:
Since passing "false" as the second argument to Symfony\Component\HttpKernel\Kernel::getBundle() is deprecated as of 3.4 and will be removed in 4.0, I did a small version verification, to assure backwards compatibility.
I am aware of the discussion in issue #375.
 
Is this solution good enough for your standards?

### DIFF
--- a/tests/Provider/BuilderAliasProviderTest.php
+++ b/tests/Provider/BuilderAliasProviderTest.php
@@ -6,6 +6,7 @@ use Knp\Bundle\MenuBundle\Provider\BuilderAliasProvider;
 use Knp\Bundle\MenuBundle\Tests\Stubs\TestKernel;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\HttpKernel\Bundle\BundleInterface;
+use Symfony\Component\HttpKernel\Kernel;
 
 class BuilderAliasProviderTest extends TestCase
 {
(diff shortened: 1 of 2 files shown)


ricardocorreia:
For some reason this change makes travis to fail on symfony 3 and php 7.1.
Does anyone have a clue why?

tests/Provider/BuilderAliasProviderTest.php
```diff
@@ -218,11 +219,21 @@ private function createMockKernelForStub()
...
-            ->with('FooBundle', false)
-            ->will($this->returnValue(array($bundle)))
-        ;
+        // Fix deprecation notice on Symfony 3.4 and higher, assuring BC
+        $version = Kernel::MAJOR_VERSION.Kernel::MINOR_VERSION;
+        if ($version >= BuilderAliasProvider::VERSION_DEPRECATION) {
+            $kernel->expects($this->once())
+                ->method('getBundle')
+                ->with('FooBundle')
+                ->will($this->returnValue(array($bundle)))
```
DonCallisto: This will return not the array but the `$bundle` as object. However that's not what causing the error. I'm trying to figure out what's the cause.

DonCallisto:
I've done this with https://github.com/DonCallisto/KnpMenuBundle/pull/1 following the "second path" listed above.
Still miss the docs part that should be part of another PR.
WDYT?

DonCallisto:
@stof please, could you (or ping someone else that can do this) take a look at this solution?
Moreover I've found that's not strictly necessary to implement my solution as second parameter (that now cause the deprecation warning) will be just ignored.
BTW even if it's not tecnically required, it should be removed and we should advice sf < 4 users about the new version of this bundle (see previous comments)

DonCallisto:
No feedbacks yet?

dbu:
this has been fixed in d1e8252829b3d0cf9a2af7ca09f04cce908cb4c8 (php ignores the superflous extra parameter in older symfony versions).

okay if we close this?

DonCallisto:
I suppose that's fine.

dbu:
thanks. and sorry for the long wait. trying to clean up a bit so that we can do a release ;-)

DonCallisto:
No probs :smile:

PR #389, opened 2018-05-24
Array syntax conversion: short syntax adopted

DonCallisto:


DonCallisto:
Please, if any PR will be merged *before* this, let me know as I can adapt accordingly.
Otherwise, if any of PRs open that will be merged *after* this, check if no `array()` is present in them.

DonCallisto:
> i would recommend to add a styleci integration in this bundle to automate this

I strongly agree

dbu:
@stof i don't have admin rights on this repo - would you mind adding styleci support in the settings? the config file could look something like this one: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/.styleci.yml

PR #390, opened 2018-08-23
Update index.rst

patrickmaynard:
Deleted two apostrophes that were misapplied to simple plurals. (Apostrophes are generally used with possessives but not with simple plurals.)


dbu:
thank you!

PR #392, opened 2018-09-16
Implement a MenuBuilderProviderInterface to support autoconfiguration

stof:
This interface allows defining the menu names from the class itself instead of having to define them in the service definition. This is similar to the event subscriber vs event listener system in Symfony.
The benefit of this interface is that the service definition can rely on autoconfiguration to register everything, which fits very well with the Flex way.

TODOs:
- [ ] write tests

### DIFF
--- a/src/DependencyInjection/Compiler/MenuBuilderPass.php
+++ b/src/DependencyInjection/Compiler/MenuBuilderPass.php
@@ -2,6 +2,7 @@
 
 namespace Knp\Bundle\MenuBundle\DependencyInjection\Compiler;
 
+use Knp\Bundle\MenuBundle\MenuBuilderProviderInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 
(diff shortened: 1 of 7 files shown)


src/DependencyInjection/Compiler/MenuBuilderPass.php
```diff
@@ -44,6 +45,38 @@ public function process(ContainerBuilder $container)
...
+
+            if (!$def->isPublic()) {
+                throw new \InvalidArgumentException(sprintf('Menu builder services must be public but "%s" is a private service.', $id));
+            }
+
+            if ($def->isAbstract()) {
+                throw new \InvalidArgumentException(sprintf('Abstract services cannot be registered as menu builders but "%s" is.', $id));
+            }
+
+            // We must assume that the class value has been correctly filled, even if the service is created by a factory
```
dbu: why must we assume that the `class` is configured if this is a factory created service?

src/Resources/doc/menu_convention.rst
```diff
@@ -0,0 +1,110 @@
+Convention-based menus
```
dbu: is there additional value in keeping this method? if not, i would deprecate the mode and drop it in the next major version. i feel we have too many similar ways to achieve the same with this bundle.

stof: I'm totally in favor of deprecating this provider (especially as it is not compatible with Flex), but this does not belong to this PR

src/DependencyInjection/Compiler/MenuBuilderPass.php
stof: because we need to call the static method on the class.

Note that Symfony's EventSubscriber does the same (this comment is a copy/paste of the Symfony one actually)

dbu: ah right, the `$class::` below... 
ok, now i get it.

garak:
@stof did you make any progress with tests here? Do you need some help with it?

stof:
No progress, because I totally forgot that PR.

craigh:
hoping this PR will get merged and a new release is on the horizon 🌤

garak:
Ehi @stof, sometimes I come back here to bother you about possible finalization of this old PR 😉
Any news?

stof:
Rather than this MenuBuilderProviderInterface providing a static method returning a map of menu name to builder methods, I think a `AsMenuBuilder` attribute applied on methods will be better (and Symfony 5 supports that for autoconfiguration)

Issue #395, opened 2018-11-30
Compatibility with Symfony 4.2


QuentinCurtet:
Upgrading to symfony 4.2 with this bundle leads to this deprecation :
> A tree builder without a root node is deprecated since Symfony 4.2 and will not be supported anymore in 5.0.

It is precisely in this file : 
https://github.com/KnpLabs/KnpMenuBundle/blob/master/src/DependencyInjection/Configuration.php

althaus:
New version'd be:

~~~php
    $treeBuilder = new TreeBuilder('knp_menu');
    $rootNode = $treeBuilder->getRootNode();
~~~

I guess this'd need a Symfony version check to use one or the other style.

QuentinCurtet:
Yes i think
I don't know if it could cause BC breaks.

But i think this could be interesting to PR your solution, can you do this ?

umulmrum:
I built a possible solution in https://github.com/KnpLabs/KnpMenuBundle/pull/396

garak:
This has been fixed with merge of #399 
This issue should be closed

PR #396, opened 2018-12-08
Adjust to Symfony 4.2 config tree builder

umulmrum:
Proposed solution for #395

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -19,7 +19,7 @@ class Configuration implements ConfigurationInterface
      */
     public function getConfigTreeBuilder()
     {
-        if (\method_exists(TreeBuilder::class, '__construct')) {
+        if (\method_exists(TreeBuilder::class, 'getRootNode')) {
             $treeBuilder = new TreeBuilder('knp_menu');
             $rootNode = $treeBuilder->getRootNode();
         } else {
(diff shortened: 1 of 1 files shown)


src/DependencyInjection/Configuration.php
```diff
@@ -19,8 +19,13 @@ class Configuration implements ConfigurationInterface
      */
     public function getConfigTreeBuilder()
     {
-        $treeBuilder = new TreeBuilder();
-        $rootNode = $treeBuilder->root('knp_menu');
+        if (\method_exists(TreeBuilder::class, '__construct')) {
```
stof: The detection should be based on `getRootNode` instead.

And I would always call `$treeBuilder = new TreeBuilder('knp_menu');`, as done in many other bundles. Passing the extra argument on older Symfony versions is not an issue.

umulmrum: OK, checking for getRootNode() now.

I saw the constructor call in other bundles, but that looks quite smelly - let me know if you insist on this, otherwise I'd keep the cleaner distinction.

althaus:
Any news on this? Does the code still need to be changed?

Surf-N-Code:
What's the status on this?
Thanks!

GuillaumeSTEIN:
ok go ?

althaus:
@stof You still need your requested changes? This seems to block the PR. The alternate PR #399 is also not merged yet.

schnoog:
Would be great to see this patched.

garak:
Since #399 was merged, this PR should be closed

dbu:
fixed in #399

PR #398, opened 2019-01-05
Bumped PHP version to 7.1

DonCallisto:


DonCallisto:
Shouldn't we also bump all `symfony/*` stuff to `2.8` and `3.4`?

dbu:
i think for the next minor version we should bump the symfony versions to those supported according to the [roadmap](https://symfony.com/roadmap). i 'd go with ^2.8|^3.4|^4.2. do you agree @stof ?

DonCallisto:
@dbu 2.8 is security fixes only ATM.

dbu:
yeah, could also drop that to make life easier. people stuck on 2.8 could still use version 2.1 of this bundle.

@stof what would you say get this PR merged with symfony version restrictions and then tag 2.2?

composer.json
```diff
@@ -19,14 +19,14 @@
         }
     ],
     "require":      {
-        "php": "^5.6 || ^7",
+        "php": "^7.1",
         "knplabs/knp-menu": "~2.3",
-        "symfony/framework-bundle": "~2.7|~3.0 | ^4.0"
+        "symfony/framework-bundle": "~2.7 | ~3.0 | ^4.0"
```
garak: Can you change this (and other relevant lines) to `^3.4 || ^4.0`?

DonCallisto: @garak done. Do you think I should also bump php version to 7.2 as 7.1 has ended its active support, or as it still receives security support is ok to leave at it is?
I'll vote for the first but I have really little experience in this kind of decisions when talking about OSS.

DonCallisto: I've also rebased on `master` in order to make all the tests pass

garak: @DonCallisto I see little advantage in excluding 7.1, we can think to raise requirements more in a next release

.travis.yml
```diff
@@ -13,22 +13,16 @@ env:
...
     - 7.1
     - 7.2
     - 7.3
     - nightly
 
 matrix:
     include:
-        - php: 5.6
-          env: COMPOSER_FLAGS='--prefer-lowest --prefer-stable' SYMFONY_DEPRECATIONS_HELPER=weak
         - php: 7.1
```
garak: I guess that, now that 7.1 is the minimum version, we should move old 5.6 options here

dbu: yep, one build should be with `--prefer-lowest` to detect when new code introduces incompatibilities with old versions (so that its either fixed or minimum requirements are increased)

DonCallisto: Ok so would you add another matrix entry leaving also `dev` dependencies and `SYMFONY_VERSION` (basically having three entries with PHP7.1 as version)?

DonCallisto: Moreover should I add an `allow_failures` with the nightly build? I think we should as PHP is moving faster than this bundle does.

dbu: afaik nightly is now php 8 which makes no sense to test as its not supported by anything yet... i'd just remove that from the matrix.

the 3 entries with php 7.1 all do different things. though i guess the idea of the one with `dev` was to use the newest supported php version, so it could be changed to 7.3.

garak: Maybe 7.4snapshot makes sense?

dbu: yeah that would make sense - could be allowed to fail maybe, as its not necessarily working correctly.

DonCallisto: Fix pushed. Does it LGTY?
Moreover with PHP7.3 and dev deps we found a potential issue with (I suppose) `phpunit-bridge` --> https://travis-ci.org/DonCallisto/KnpMenuBundle/jobs/548106697 but is not related with this issue. If everything is good with this PR I can also fix that problem in a different PR.

DonCallisto:
Also fixed tests in https://github.com/KnpLabs/KnpMenuBundle/pull/406

DonCallisto:
I just noticed that the fix applied in above mentioned PR is only fine as long as it is included in this PR.
What you guys think? 
It's ok to cherry pick the commit and make those two a whole PR?

PR #400, opened 2019-03-02
Added parameters for services

libertjeremy:
Here is a PR to add parameters to define the class of services (LazyProvider / BuilderServiceProvider), just as the other services were already defined :
- https://github.com/KnpLabs/KnpMenuBundle/blob/v2.2.1/src/Resources/config/menu.xml#L50
- https://github.com/KnpLabs/KnpMenuBundle/blob/v2.2.1/src/Resources/config/menu.xml#L60
- ...

### DIFF
--- a/src/Resources/config/menu.xml
+++ b/src/Resources/config/menu.xml
@@ -10,8 +10,10 @@
         <parameter key="knp_menu.helper.class">Knp\Menu\Twig\Helper</parameter>
         <parameter key="knp_menu.matcher.class">Knp\Menu\Matcher\Matcher</parameter>
         <parameter key="knp_menu.menu_provider.chain.class">Knp\Menu\Provider\ChainProvider</parameter>
+        <parameter key="knp_menu.menu_provider.lazy.class">Knp\Menu\Provider\LazyProvider</parameter>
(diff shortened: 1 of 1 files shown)


src/Resources/config/menu.xml
```diff
@@ -44,7 +46,7 @@
             <argument type="collection" />
         </service>
 
-        <service id="knp_menu.menu_provider.lazy" class="Knp\Menu\Provider\LazyProvider" public="false">
+        <service id="knp_menu.menu_provider.lazy" class="%knp_menu.menu_provider.lazy.class%" public="false">
```
dbu: according to the updated symfony best practices, this should no longer be done. if you want to change how a service is defined, define one with the same name in your application, or write a compiler pass to change all necessary things.

afaik the existing parameters are only here for BC and will be removed in the next major version.

dbu:
@stof okay if we close this?

stof:
Yeah, this has no reason to be accepted, as it goes against Symfony best practices and would make it harder to maintain BC (as changing arguments in a service definition could break things for people overriding the class, which is precisely why Symfony recommends not using parameters for the class).

PR #404, opened 2019-04-16
Add a priority to allow controlling the order of providers

emulienfou:
This Pull Request introduce a new feature to allow specifying the priority of providers during the registration of a new provider when this one has been tagged. (Similar has voters)

So, a developer will now be able to write something like:
```yaml
services:
  app.menu_provider:
        class: AppBundle\Provider\CustomMenuProvider
        arguments:
          - @knp_menu.factory
        tags:
          - { name: knp_menu.provider, priority: 20 }
```

### DIFF
--- a/tests/DependencyInjection/Compiler/AddProvidersPassTest.php
+++ b/tests/DependencyInjection/Compiler/AddProvidersPassTest.php
@@ -2,7 +2,10 @@
 
 namespace Knp\Bundle\MenuBundle\Tests\DependencyInjection\Compiler;
 
+use function call_user_func_array;
+use function class_exists;
 use Knp\Bundle\MenuBundle\DependencyInjection\Compiler\AddProvidersPass;
+use function krsort;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\DependencyInjection\Argument\IteratorArgument;
 use Symfony\Component\DependencyInjection\Reference;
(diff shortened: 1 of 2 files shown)


tests/DependencyInjection/Compiler/AddProvidersPassTest.php
```diff
@@ -84,4 +87,50 @@ public function testProcessForManyProviders()
...
+
+        krsort($expectedProviders);
+        $expectedProviders = call_user_func_array('array_merge', $expectedProviders);
+
+        if (class_exists(IteratorArgument::class)) {
+            $expectedProviders = new IteratorArgument($expectedProviders);
+        }
+
+        $definitionMock = $this->getMockBuilder('Symfony\Component\DependencyInjection\Definition')
+            ->disableOriginalConstructor()
```
dbu: you can use `$this->createMock(Class)` instead of getMockBuilder->disableOriginalConstructor->getMock (at least i think so, not entirely sure if PHP 5.6 will install a too old version of phpunit that is missing this feature.

tests/DependencyInjection/Compiler/AddProvidersPassTest.php
```diff
@@ -84,4 +87,50 @@ public function testProcessForManyProviders()
...
+            ->getMock();
+        $definitionMock->expects($this->once())
+            ->method('replaceArgument')
+            ->with($this->equalTo(0), $expectedProviders);
+
+        $containerBuilderMock = $this->getMockBuilder('Symfony\Component\DependencyInjection\ContainerBuilder')->getMock();
```
dbu: createMock

tests/DependencyInjection/Compiler/AddProvidersPassTest.php
```diff
@@ -84,4 +87,50 @@ public function testProcessForManyProviders()
...
     }
+
+    public function testPriorityRegisteredProviders()
+    {
+        $expectedProviders = [];
+        $expectedProviders[0][] = new Reference('id');
+        $expectedProviders[20][] = new Reference('id2');
+
+        krsort($expectedProviders);
+        $expectedProviders = call_user_func_array('array_merge', $expectedProviders);
```
dbu: you should craft the expected list directly rather than have the same code as the class that is tested.

tests/DependencyInjection/Compiler/AddProvidersPassTest.php
```diff
@@ -84,4 +87,50 @@ public function testProcessForManyProviders()
...
+
+        $containerBuilderMock = $this->getMockBuilder('Symfony\Component\DependencyInjection\ContainerBuilder')->getMock();
+        $containerBuilderMock->expects($this->once())
+            ->method('hasDefinition')
+            ->will($this->returnValue(true));
+        $containerBuilderMock->expects($this->once())
+            ->method('findTaggedServiceIds')
+            ->with($this->equalTo('knp_menu.provider'))
+            ->will($this->returnValue([
+                'id2' => ['provider_tag2', ['priority' => 20]],
```
dbu: can you add the example with multiple tags here, to cover that edge case too? and use -20 for this provider, so the krsort actually does reverse something?

tests/DependencyInjection/Compiler/AddProvidersPassTest.php
emulienfou: Hi @dbu didn't understand exactly what you mean by that?

GawainLynch: @emulienfou just simply that the "expected" value(s) should be as literal as possible, in this case (assuming I've had enough :coffee:):
```
$expectedProviders = [
    new Reference('id'),
    new Reference('id2'),
];
```

That way you're not testing against what we *think* might happen with, in this case, the `ksort()` and `marray_merge()`, but rather something that's easy to visually verify.

dbu: exactly ;-)

emulienfou: Oh off course thank, really needed a coffee lol

tests/DependencyInjection/Compiler/AddProvidersPassTest.php
```diff
@@ -84,4 +88,43 @@ public function testProcessForManyProviders()
...
+
+    public function testPriorityRegisteredProviders()
+    {
+        $expectedProviders = [new Reference('id2'), new Reference('id')];
+
+        if (class_exists(IteratorArgument::class)) {
+            $expectedProviders = new IteratorArgument($expectedProviders);
+        }
+
+        $definitionMock = $this->createMock('Symfony\Component\DependencyInjection\Definition');
```
dbu: please use `Definition::class` for this, with a `use` statement for the Definition class. same for the mocking below

tests/DependencyInjection/Compiler/AddProvidersPassTest.php
```diff
@@ -6,9 +6,11 @@
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\DependencyInjection\Argument\IteratorArgument;
 use Symfony\Component\DependencyInjection\Reference;
+use function class_exists;
 
 class AddProvidersPassTest extends TestCase
 {
+
```
dbu: please revert the formatting changes to this file that are not related to the feature you add (this one is wrong. the newline things would not be necessary if createMock with the ::class would be used, but such cleanup should not be mixed into a new feature)

garak:
I think this PR should include an update to documentation

PR #412, opened 2019-09-06
version 3 (compatible with KnpMenu v3)

garak:


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 3.0.0 (to be released)
+
+* [BC] removed deprecated features
```
stof: should be `[BC break]`, not `[BC]` (the part making it worth highlighting it is precisely `break`)

src/DependencyInjection/Compiler/RegisterMenusPass.php
```diff
@@ -22,16 +22,7 @@ public function process(ContainerBuilder $container)
...
-        // so the older providers will be used.
-        if (!class_exists(ServiceClosureArgument::class)) {
-            $container->removeDefinition('knp_menu.menu_provider.lazy');
-
-            return;
-        }
-
         // Remove the old way of handling this feature.
-        $container->removeDefinition('knp_menu.menu_provider.container_aware');
         $container->removeDefinition('knp_menu.menu_provider.builder_service');
```
stof: you should drop this service definition entirely now, as it is always useless when the LazyProvider is used.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
...
+* [BC] made final all classes that implement an interface
+* Dropped support for KnpMenu 2
+* Increased minimum PHP version to 7.2 (to be consistent with KnpMenu 3)
```
stof: The fact that class parameters are removed in the container should also be documented.

src/Resources/config/menu.xml
```diff
@@ -49,7 +38,7 @@
             <tag name="knp_menu.provider" />
         </service>
 
-        <service id="knp_menu.menu_provider.container_aware" class="%knp_menu.menu_provider.container_aware.class%" public="false">
+        <service id="knp_menu.menu_provider.container_aware" class="Knp\Menu\Provider\PsrProvider" public="false">
```
stof: The service definition needs to be dropped, as the compiler pass is removing it anyway on modern Symfony versions using LazyProvider (or at least, it was before this PR).

garak: Are you sure? I'm trying this in a project of mine (with Symfony 3.4) and I can see `knp_menu.menu_provider.container_aware` service in `debug:container`

stof: are you trying your own branch, or the existing master branch ? Your own branch does not remove it anymore, because you deleted the line removing the definition.

garak: So, if I understand correctly, we should remove the whole `MenuPass` too

garak: @stof can we also remove `container_aware` option?

src/DependencyInjection/Compiler/AddRenderersPass.php
```diff
@@ -36,16 +36,10 @@ public function process(ContainerBuilder $container)
...
-            $container->register('knp_menu.renderer_provider', PsrProvider::class)
-                ->addArgument($locator)
-                ->addArgument('%knp_menu.default_renderer%');
-        } else {
-            // BC for Symfony < 3.3
-            $definition = $container->getDefinition('knp_menu.renderer_provider');
-            $definition->replaceArgument(2, $renderers);
-        }
+        $locator = ServiceLocatorTagPass::register($container, $rendererReferences);
+        // Replace the service definition with a PsrProvider
```
stof: the replacement itself should be dropped. The service should be a PsrProvider directly in the XML config file now, and we should only replace the first argument with the locator here.

src/DependencyInjection/Compiler/AddVotersPass.php
```diff
@@ -23,7 +23,6 @@ public function process(ContainerBuilder $container)
         }
 
         $definition = $container->getDefinition('knp_menu.matcher');
-        $listener = $container->getDefinition('knp_menu.listener.voters');
 
         $hasRequestAwareVoter = false;
```
stof: should be dropped now (as well as the following code using it)

src/Resources/config/menu.xml
stof: yes. MenuPass should dropped in favor of RegisterMenuPass

And the option should also be dropped (deprecating it in 2.x first)

tests/DependencyInjection/Compiler/MenuBuilderPassTest.php
```diff
@@ -40,47 +40,43 @@ public function testNoopWithoutProvider()
...
     }
 
-    /**
-     * @expectedException \InvalidArgumentException
-     * @expectedExceptionMessage Menu builder services must be public but "id" is a private service.
-     */
     public function testFailsWhenServiceIsPrivate()
     {
+        $this->expectException(\InvalidArgumentException::class);
+        $this->expectExceptionMessage('Menu builder services must be public but "id" is a private service.');
```
stof: Migrating from PHPUnit annotations to PHPUnit methods for exception expectations should be done separately IMO.

Thus, the main benefit of these methods is that they can set the expectation just before the place where we expect it (so that an exception during the setup of the test does not make it pass). So these calls should not be placed at the top of the test, but just before the call to the tested code (the call to `->process()` here).

stof: btw, this should be done in the 2.x branch to make it simpler to merge branches together if needed (the new methods are available since PHPUnit 5, which can be used on PHP 7.1)

src/DependencyInjection/Compiler/AddRenderersPass.php
garak: @stof in a previous comment you said that definition of `Knp\Menu\Provider\PsrProvider` service should be dropped (since it's removed using LazyProvider), now you say that the service should be in XML config. Can you clarify?

stof: you are confusing things. This compiler pas is about the renderer provider, not the menu provider.

The PsrProvider for renderer provider must be used, as that's what get used for Symfony 3.3+. But the service definition using the PsrProvider should be in the XML file directly, instead of defining the service for the old ContainerAwareProvider in XML and replacing it entirely here.
Then, this compiler pass should be only replacing the argument to inject the service locator.

src/DependencyInjection/Compiler/MenuBuilderPass.php
```diff
@@ -12,9 +12,9 @@
  *
  * @internal
  */
-class MenuBuilderPass implements CompilerPassInterface
+final class MenuBuilderPass implements CompilerPassInterface
```
stof: please tag them as `@final` in a 2.x release to be safe.

src/Resources/config/menu.xml
```diff
@@ -49,41 +38,32 @@
             <tag name="knp_menu.provider" />
         </service>
 
-        <service id="knp_menu.menu_provider.container_aware" class="%knp_menu.menu_provider.container_aware.class%" public="false">
-            <argument type="service" id="service_container" />
-            <argument type="collection" />
-        </service>
-
         <service id="knp_menu.menu_provider.builder_service" class="Knp\Bundle\MenuBundle\Provider\BuilderServiceProvider" public="false">
```
stof: this service must be removed. It is useless now that the lazy provider can always be used (see the compiler pass always removing it). And the class should be removed too.

tests/DependencyInjection/KnpMenuExtensionTest.php
```diff
@@ -72,8 +70,7 @@ public function testDisableContainerAwareProvider()
     {
         $container = new ContainerBuilder();
         $loader = new KnpMenuExtension();
-        $loader->load([['providers' => ['container_aware' => false]]], $container);
+        $loader->load([['providers' => []]], $container);
```
stof: this legacy test should be removed entirely instead, as it is the test about disabling the ContainerAwareProvider

src/DependencyInjection/Compiler/AddRenderersPass.php
garak: So, should we remove ContainerAwareProvider class to?

src/DependencyInjection/Compiler/MenuBuilderPass.php
phansys: Which branch is intended to target 2.x release? Currently, there are only 2: [`v3` and `master`](https://github.com/KnpLabs/KnpMenuBundle/branches).

garak: @phansys master

phansys: Then, shouldn't this PR target `v3`?

garak: No, v3 is just a temporary branch where this PR is already merged, useful to do a practical test.

garak:
It looks to me that all notes has been addressed. I used v3 branch (that is a copy of this PR) on a real project without problems.
@stof do you think we can merge and release?

PR #413, opened 2019-09-06
remove deprecated annotations

garak:


stof:
Would be good to take the other part of my comment (https://github.com/KnpLabs/KnpMenuBundle/pull/412#discussion_r321779758) into account too.

tests/DependencyInjection/Compiler/AddExtensionsPassTest.php
```diff
@@ -102,6 +100,8 @@ public function testMissingAddExtension()
             ->method('getParameterBag')
             ->willReturn($parameterBagMock);
 
+        $this->expectException(\Symfony\Component\Config\Definition\Exception\InvalidConfigurationException::class);
```
derrabus: That's probably a matter of taste, but we could now import the class as well.

stof: indeed, please do it.

garak: Done!

PR #414, opened 2019-09-06
deprecate "container_aware" option

garak:


src/DependencyInjection/Configuration.php
```diff
@@ -34,7 +34,7 @@ public function getConfigTreeBuilder()
                     ->addDefaultsIfNotSet()
                     ->children()
                         ->booleanNode('builder_alias')->defaultTrue()->end()
-                        ->booleanNode('container_aware')->defaultTrue()->end()
+                        ->booleanNode('container_aware')->setDeprecated('The "%node%" option is deprecated and will be removed in version 3.')->defaultTrue()->end()
```
stof: is this feature available in all our target symfony/config versions ? Or do we need to trigger the deprecation manually in `beforeNormalization` ?

src/DependencyInjection/Configuration.php
```diff
@@ -34,7 +34,7 @@ public function getConfigTreeBuilder()
...
                         ->booleanNode('builder_service')->defaultTrue()->end()
```
stof: that one should be deprecated for the same reason (LazyProvider also makes this service obsolete)

src/DependencyInjection/Configuration.php
garak: Feature has been [introduced in 3.4](https://symfony.com/blog/new-in-symfony-3-4-deprecate-configuration-options) (that is the minimum Symfony version we support)

derrabus: So, in other words, version 3 will behave like both settings have been set to their default values?

stof: These 2 settings are here to allow unregistered these 2 providers. But when running on Symfony 3.4+, they are both obsoleted by the LazyProvider, and so get unregistered 100% of the time.

Thus, these 2 settings should never have been needed. The ContainerAwareProvider and the BuilderServiceProvider are both configured by compiler passes. So we know when building the container whether they are useful at runtime or no (if they don't contain any menu definition, they are useless). This is the reason why the LazyProvider does not have an equivalent config option (everything happens automatically).
The config option is useful for `builder_alias`, because it is convention-based, not configuration-based, and so we don't know the list of menus it can build (and so being able to manually disable it if your project does not use it is good).

.travis.yml
```diff
@@ -11,6 +11,7 @@ env:
     global:
         - SYMFONY_PHPUNIT_DIR="$HOME/symfony-bridge/.phpunit"
         - SYMFONY_PHPUNIT_REMOVE="symfony/yaml" # keep Prophecy
+        - SYMFONY_DEPRECATIONS_HELPER=2 # we need to allow our own deprecations
```
stof: we should not do that. We should mark the test triggering them as legacy instead.

garak:
Requested changes should be implemented. Failure on dev in travis should be unrelated.
Can we merge?

src/DependencyInjection/KnpMenuExtension.php
```diff
@@ -52,15 +52,15 @@ public function load(array $configs, ContainerBuilder $container)
     /**
      * {@inheritdoc}
      */
-    public function getNamespace()
+    public function getNamespace(): string
```
stof: this should be done as a separate PR IMO, as it is a totally separate change (a valid one, but unrelated to deprecating these options)

garak: I did it because that missing return type is generating a deprecation on latest versions of deps

garak: So, it was just an attempt to avoid failing test. But I see that test is still failing, for unrelated changes (that are object of another PR), so I guess we can accept such failure

PR #416, opened 2019-10-24
Support for Symfony 5

GawainLynch:
- Replace deprecated PHPUnit @expectedException annotations
- Update deprecated use of HttpFoundation events
- Fix Travis config

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: php
 sudo: false
 
 cache:
-    directory:
+    directories:
         - $HOME/.composer/cache/files
         - $HOME/symfony-bridge/.phpunit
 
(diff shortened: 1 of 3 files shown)


garak:
@GawainLynch it's merged now.
You can rebase
Thanks

GawainLynch:
Rebased, @garak. I'll keep an eye on Travis, just in case.

.travis.yml
```diff
@@ -20,10 +20,13 @@ php:
...
+              - SYMFONY_DEPRECATIONS_HELPER=weak
         - php: 7.3
-          env: DEPENDENCIES='dev'
-          env: SYMFONY_DEPRECATIONS_HELPER='max[self]=0'
+          env:
+              - DEPENDENCIES='dev'
+              - SYMFONY_DEPRECATIONS_HELPER='max[self]=0'
         # Test LTS version
         - php: 7.3
           env: SYMFONY_VERSION='^3'
```
garak: I guess we should add Symfony 5 to our matrix (and maybe use DEPENDENCIES dev on that)
Incidentally, we're missing Symfony 4 too

GawainLynch: Also, the version of PHPUnit used by `simple-phpunit` had to be pinned at 7.5. Travis didn't report a fail, but pretty sure we don't want [this](https://travis-ci.org/KnpLabs/KnpMenuBundle/jobs/614518051)

phpunit.xml.dist
```diff
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <phpunit colors="true" bootstrap="vendor/autoload.php">
+    <server name="SYMFONY_PHPUNIT_VERSION" value="7.5" />
```
garak: I'd rather use 8.4, that's latest phpunit

GawainLynch: WFM :+1:

.travis.yml
```diff
@@ -20,13 +20,20 @@ php:
...
+        - php: 7.3
+          env: SYMFONY_VERSION='^4@dev'
+        - php: 7.3
+          env: SYMFONY_VERSION='^5@dev'
```
garak: You should use `- DEPENDENCIES='dev'` here, and remove it from line 28, where doesn't make sense.
Moreover, we don't need dev deps on Symfony 4 (I know there's 4.4-beta now, but stable will be out in a few days)

GawainLynch: > Moreover, we don't need dev deps on Symfony 4 (I know there's 4.4-beta now, but stable will be out in a few days)

Happy to go with just `^4`, but until (roughly) release this error will be thrown:
```
The requested package symfony/lts ^4 is satisfiable by symfony/lts[4.x-dev] but these conflict with your requirements or minimum-stability
```

Is that the prefered option?

garak: We need to remove symfony/lts, it's deprecated. To force Symfony version we can rely on requiring related `symfony/framework-bundle` package, or replace version with sed. For this last option, see https://github.com/PUGX/filter-bundle/blob/master/.travis.yml#L37

GawainLynch: After a couple of facepalm-worthy pushes, I think I have it now :smile_cat:

garak:
Thank you!

PR #431, opened 2020-07-15
Run tests against a real ContainerBuilder

derrabus:
The testsuite currently fails when run against Symfony 5.1 because of internal changes in Symfony's `ContainerBuilder`.

The failing tests mock the `ContainerBuilder` class. This PR removes those mocks, which should render the tests more robust and readable.

### DIFF
--- a/tests/DependencyInjection/Compiler/AddExtensionsPassTest.php
+++ b/tests/DependencyInjection/Compiler/AddExtensionsPassTest.php
@@ -3,119 +3,57 @@
 namespace Knp\Bundle\MenuBundle\Tests\DependencyInjection\Compiler;
 
 use Knp\Bundle\MenuBundle\DependencyInjection\Compiler\AddExtensionsPass;
-use Knp\Menu\FactoryInterface;
-use Knp\Menu\ItemInterface;
+use Knp\Menu\MenuFactory;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Reference;
 
 class AddExtensionsPassTest extends TestCase
 {
     public function testProcessWithoutProviderDefinition()
     {
-        $containerBuilder = $this->getMockBuilder('Symfony\Component\DependencyInjection\ContainerBuilder')->getMock();
-        $containerBuilder->expects($this->once())
-            ->method('has')
-            ->willReturn(false);
-        $containerBuilder->expects($this->never())
-            ->method('findTaggedServiceIds');
+        $containerBuilder = new ContainerBuilder();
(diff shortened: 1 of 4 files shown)


garak:
While I really appreciate such effort, I'm not really convinced of the opportunity to replace a mock with a real instance in our tests.
I'm not a testing expert indeed, but I know that tests should only create instances of SUT, while mocking any external deps.
Do you think we could solve this specific problem by using matthiasnoback/symfony-dependency-injection-test instead?

derrabus:
> While I really appreciate such effort, I'm not really convinced of the opportunity to replace a mock with a real instance in our tests.

The real instance is a builder, an object that has the purpose of mutate a complex data structure. Mocking such a class in a robust way is close to impossible, because you need to make assumptions about internal behavior of that class. Your tests will break everytime that behavior changes, as we can observe on Travis at the moment.

If you use the real builder, you can test if the interaction between your compiler pass and the builder actually produces the desired data structure.

> I'm not a testing expert indeed, but I know that tests should only create instances of SUT, while mocking any external deps.

And what is your SUT in this case? What do you want to test?

A compiler pass is totally useless on its own. It only does useful things as an extension to the container builder. And you cannot test an extension properly without the system that it extends. Your SUT should be the combination of the pass and the data structure that it's supposed to mutate.

In other words: Do you want to make sure that your compiler pass calls certain methods in the right order or do you want to test if your compiler pass produces the desired dependency tree?

> Do you think we could solve this specific problem by using matthiasnoback/symfony-dependency-injection-test instead?

That library will almost certainly work, but it also does not mock the builder. Under the hood, it pretty much does what my tests do as well. 😃 

Don't get me wrong. Mocks can be a powerful tool, but I wouldn't use them on builder objects and pure data containers.

garak:
OK, my concern with the current proposed solution is that it looks like a bit complex, and I'd like to move such complexity outside (because maybe we'll have to replicate it in a different tests or in a different bundle).
So, I'd love to come to an elegant and sharable solution, this is why I thought to Noback's library.
Do you think that it could be better? If not, I'm OK with current solution.

derrabus:
I mean, the tests aren't that complex, really. The diff might look a bit scary, though. Maybe have a look at the code without the diff. It's always the same structure:

* Setup the initial dependency tree.
* Run the compiler pass.
* Run assertions on the produced dependency tree.

All that Noback's library does is creating abstract test case classes around that procedure. Those classes just wrap calls to the container builder. So `$builder->register('my_service', SomeClass::class)` becomes `$this->addDefinition('my_service', new Definition(SomeClass::class))`.

As far as I can tell, the test would not become less complex by using that library. I wouldn't use it here.

garak:
Hey @derrabus do you mind to propose same fix on KnpPaginatorBundle? Thanks

derrabus:
I'll have a look.

derrabus:
KnpLabs/KnpPaginatorBundle#649

PR #432, opened 2020-07-15
Use a variadic call instead of call_user_func_array()

derrabus:
This is a minor performance improvement. Calling `array_merge()` directly should be slightly faster than using `call_user_func_array()` for that matter.


stof:
AFAIK, a qualified call to `call_user_func_array` has no overhead, as it is optimized at compile-time. But I agree with the array spreading (this `...` is not about variadic or no. The call was already using the variadic signature of `array_merge` before)

PR #436, opened 2020-11-02
add more coding standard

garak:


src/DependencyInjection/Compiler/AddExtensionsPass.php
```diff
@@ -30,11 +30,9 @@ public function process(ContainerBuilder $container): void
 
         $definition = $container->findDefinition('knp_menu.factory');
 
-        if (!method_exists($container->getParameterBag()->resolveValue($definition->getClass()), 'addExtension')) {
-            throw new InvalidConfigurationException(sprintf(
-                'To use factory extensions, the service of class "%s" registered as knp_menu.factory must implement the "addExtension" method',
-                $definition->getClass()
-            ));
+        if (!\method_exists($container->getParameterBag()->resolveValue($definition->getClass()), 'addExtension')) {
+            $msg = 'To use factory extensions, the service of class "%s" registered as knp_menu.factory must implement the "addExtension" method';
```
stof, line 34: I'd rather keep the pattern inline in the `sprintf` call, which IDEs validate as having the right number of placehilders.

.php_cs
```diff
@@ -11,7 +11,12 @@ return PhpCsFixer\Config::create()
     ->setRules([
         '@Symfony' => true,
         '@Symfony:risky' => true,
+        '@PHP71Migration:risky' => true,
+        '@PHPUnit75Migration:risky' => true,
         'array_syntax' => ['syntax' => 'short'],
+        'ordered_imports' => true,
+        'declare_strict_types' => false,
+        'native_function_invocation' => true,
```
stof, line 19: Do we want to namespace them all, or only the ones for which it makes a difference for performance (which is the config included in `@Symfony:risky` ? The extra `\` in front of all calls adds visual noise. That's the reason for the rule selected by Symfony.

src/DependencyInjection/Compiler/AddExtensionsPass.php
garak: Problem is that line is really too long. We're way over recommended 80 columns

.php_cs
garak: I'm not a big fan of micro-optimization.
I think it's better to keep it consistent (adding front slash to everything or to nothing).
If we start to make a distinction, at every change we'll be forced to check whether the slash is worth or not. That's a burden.

PR #438, opened 2021-01-14
move CI to Github Actions

garak:


.github/workflows/build.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+on:
+    pull_request: ~
+    push: ~
+
+jobs:
+    tests:
+        runs-on: ubuntu-20.04
+        strategy:
+            max-parallel: 1
```
stof, line 11: why ?

.github/workflows/build.yaml
```diff
@@ -0,0 +1,49 @@
...
+            matrix:
+                php:
+                    - '7.2'
+                    - '7.3'
+                    - '7.4'
+                    - '8.0'
```
stof, line 17: the normal jobs must have a value set for `description` or `composer_option`. Otherwise, they are swallowed by included jobs which are a super-set of their values.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,49 @@
+name: Continuous Integration
+
+on:
+    pull_request: ~
+    push: ~
+
+jobs:
+    tests:
+        runs-on: ubuntu-20.04
+        strategy:
```
stof, line 10: please add `fail-fast: false` to ensure that all jobs of the matrix are always executed even when one of them fails.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,49 @@
...
+                include:
+                    - description: 'Symfony 3.4'
+                      php: '7.2'
+                      composer_option: '--prefer-lowest'
+                    - description: 'Symfony 4.4'
+                      php: '7.3'
+                      symfony: 4.4.*
+                    - description: 'Symfony 5.1'
+                      php: '7.4'
+                      symfony: 5.1.*
```
stof, line 27: should be removed. Jobs forcing a precise version of Symfony should be used only for LTS versions IMO, so that this needs to be maintained every 2 years rather than every 6 months when versions become EOL.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,49 @@
...
+        strategy:
+            max-parallel: 1
+            matrix:
+                php:
+                    - '7.2'
+                    - '7.3'
+                    - '7.4'
+                    - '8.0'
+                include:
+                    - description: 'Symfony 3.4'
```
stof, line 19: Wrong description. this job is testing lowest deps, which is not the same than a job testing symfony 3.4 (which would rather check an uptodate 3.4 version)

.github/workflows/build.yaml
```diff
@@ -0,0 +1,49 @@
...
+                      symfony: 5.2.*
+        name: PHP ${{ matrix.php }} tests (${{ matrix.description }})
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v2
+            - name: Cache
+              uses: actions/cache@v2
+              with: 
+                  path: ~/.composer/cache/files
+                  key: ${{ matrix.php }}-${{ matrix.symfony }}-${{ matrix.composer_option }}
```
stof, line 39: I would add a fixed `composer-` prefix in front of it, in case cache want to be used for something else than the composer cache in the future

.github/workflows/build.yaml
```diff
@@ -0,0 +1,49 @@
...
+            matrix:
+                php:
+                    - '7.2'
+                    - '7.3'
+                    - '7.4'
+                    - '8.0'
+                include:
+                    - description: 'Symfony 3.4'
+                      php: '7.2'
+                      composer_option: '--prefer-lowest'
```
stof, line 21: the `prefer-lowest` job should set `SYMFONY_DEPRECATIONS_HELPER=weak` in the environment

.github/workflows/build.yaml
```diff
@@ -0,0 +1,49 @@
...
+        runs-on: ubuntu-20.04
+        strategy:
+            max-parallel: 1
+            matrix:
+                php:
+                    - '7.2'
+                    - '7.3'
+                    - '7.4'
+                    - '8.0'
+                include:
```
stof, line 18: Travis also included a job running with `minimum-stability: dev`, which got lost in the migration

.github/workflows/build.yaml
```diff
@@ -0,0 +1,49 @@
...
+                  key: ${{ matrix.php }}-${{ matrix.symfony }}-${{ matrix.composer_option }} 
+            - name: Setup PHP
+              uses: shivammathur/setup-php@v2
+              with:
+                  php-version: ${{ matrix.php }}
+            - run: |
+                  sed -ri 's/"symfony\/(.+)": "(.+)"/"symfony\/\1": "'${{ matrix.symfony }}'"/' composer.json;
+              if: matrix.symfony
+            - run: composer update --no-interaction --no-progress --ansi ${{ matrix.composer_option }}
+            - run: vendor/bin/simple-phpunit
```
stof, line 48: the `composer validate --strict --no-check-lock` test got lost. We should have it on CI (maybe with a separate job as we don't need to do it for all PHP versions and we don't even need to install deps for that)

stof, line 48: `./vendor/bin/simple-phpunit install` should be kept as a separate step so that the output of running tests does not include the installation of PHPUnit as well. running the install separately means it will be collapsed separately (and that separate step can be named `Install PHPUnit` for clarity about what it does)

stof (changes requested):

stof:
The `symfony/phpunit-bridge` constraint also needs to be updated so that the `--prefer-lowest` job works. Older releases of the bridge have been broken by the removal of old branches in the PHPUnit repo.

.github/workflows/build.yaml
garak: Are you sure? [Docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#example-including-new-combinations) say "Any unmatched include configurations are added to the matrix"

.github/workflows/build.yaml
garak: Do you see real value in it? Should we test for some features that vendors still have to release?

.github/workflows/build.yaml
garak: I'm getting problems with that. Installation of simple-phpunit depends on installation of other vendors and on environment variables, that are defined in the matrix.

.github/workflows/build.yaml
stof: yeah, but if you only add new fields which are not part of the normal matrix, this will be replace the expected normal job, because the `php: '7.2'` job will be matched by the `php: '7.2', composer_option: '--prefer-lowest` one.

stof: btw, if you look at https://github.com/garak/KnpMenuBundle/actions/runs/485202488, you'll see it only has 4 jobs in the matrix, proving my point.

.github/workflows/build.yaml
stof: this is about testing compatibility with the future release of Symfony before it is released. this allows to prepare compat so that we have a chance of being compatible with these version the day they become stable (and users use them)

.github/workflows/build.yaml
stof: well, I'm asking to put `./vendor/bin/simple-phpunit install` as a separate step just before the one running tests. It can stay after the `composer update` step.

.github/workflows/build.yaml
stof: Removing all jobs running a normal `composer update` was not the right fix for that. These jobs are important.

garak: So, should we re-add `jobs` key? With a description for each entry?

garak:
@stof do you have an hint for fatal error we're getting in phpunit with lowest deps?

>PHP Fatal error:  Declaration of SebastianBergmann\Comparator\DOMNodeComparator::assertEquals($expected, $actual, $delta = 0, $canonicalize = false, $ignoreCase = false) must be compatible with SebastianBergmann\Comparator\ObjectComparator::assertEquals($expected, $actual, $delta = 0, $canonicalize = false, $ignoreCase = false, array &$processed = Array) in /home/runner/work/KnpMenuBundle/KnpMenuBundle/vendor/sebastian/comparator/src/DOMNodeComparator.php on line 149

dbu:
not stof, but: can you narrow the phpunit version constraints? it sounds like parts of phpunit ended up in incompatible code. (maybe check what version of that comparator component got installed and if you can force that to be higher)

garak:
> not stof, but: can you narrow the phpunit version constraints? it sounds like parts of phpunit ended up in incompatible code. (maybe check what version of that comparator component got installed and if you can force that to be higher)

Nailed it: simple-phpunit is installing sebastian/comparator:3.0.3, while installed version with lower-deps is 1.1.0.
So, should we restrict versions for that dependency? Or maybe replace simple-phpunit with "real" phpunit?
Frankly, I don't see the reason to use simple-phpunit instead of phpunit (where this kind of problems are unlikely to happen)

.github/workflows/build.yaml
stof: you should re-add them, with `composer_option` using an empty string, `dev` using `false`, etc...

stof:
Well, we should either use simple-phpunit with no installation of phpunit in our vendor folder or always use the phpunit installed in our vendor folder (and using the phpunit-bridge only for its deprecation reporting features)

garak:
I hope I addressed every concern. Please re-review, thanks.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,63 @@
...
+    push: ~
+
+jobs:
+    check:
+        runs-on: ubuntu-20.04
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v2
+            - name: Composer check
+              run: composer validate --strict --no-check-lock
```
stof, line 14: you still need the `shivammathur/setup-php@v2` step before that one, otherwise composer will not be available.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,63 @@
...
+                      composer_option: '--prefer-lowest'
+                      env:
+                          SYMFONY_DEPRECATIONS_HELPER: max[self]=0
+                    - description: 'Symfony 4.4'
+                      php: '7.3'
+                      symfony: 4.4.*
+                    - description: 'Symfony 5.2'
+                      php: '7.4'
+                      symfony: 5.2.*
+                    - description: 'Symfony next'
```
stof, line 40: This should be described as `dev deps` or something like that, as it is not only about the next version of Symfony

.github/workflows/build.yaml
```diff
@@ -0,0 +1,63 @@
...
+                      php: '8.0'
+                    - description: 'Lowest deps'
+                      php: '7.2'
+                      composer_option: '--prefer-lowest'
+                      env:
+                          SYMFONY_DEPRECATIONS_HELPER: max[self]=0
+                    - description: 'Symfony 4.4'
+                      php: '7.3'
+                      symfony: 4.4.*
+                    - description: 'Symfony 5.2'
```
stof, line 37: I suggest removing this job to keep only explicit jobs for LTS versions (thus, 5.2 is currently the latest Symfony version and so already used in normal jobs).

stof: On the other hand, a job testing the latest 3.4 LTS might make sense as it is still supported in composer.json.

garak:
Anything still blocking this PR from being merged?

.github/workflows/build.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+jobs:
+    check:
+        runs-on: ubuntu-20.04
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v2
+            - name: Composer check
+              uses: shivammathur/setup-php@v2
+              run: composer validate --strict --no-check-lock
```
stof, line 15: The run mus be a separate step than the setup-php. You are missing a `-`

stof (changes requested):

.github/workflows/build.yaml
garak: Done

PR #445, opened 2021-09-20
Add Symfony 6 support

jordisala1991:

PR #446, opened 2021-09-20
Fix the CI job using dev deps

stof:
Preferring stable versions totally defeats the goal of that CI job, as it can find compatible stable versions.


jordisala1991:
That's the error I got when trying Symfony 6:

```
PHP Fatal error:  '\static' is an invalid class name in /home/runner/work/KnpMenuBundle/KnpMenuBundle/vendor/phpspec/prophecy/src/Prophecy/Doubler/Generator/ClassCreator.php(49) : eval()'d code on line 34
```

There is a mixed usage of both phpunit mock and prophecy ones. prophecy being used on 2 files only. I can replace that mocks for normal ones and get rid of that dependency if you want.

stof:
> That's the error I got when trying Symfony 6:

that's because we actually rely on a bunch of transitive dependencies right now, and so our dev job already gets some Symfony 6 packages for them.

I'll try to look whether I can fix support for `static` in prophecy instead.

garak:
It looks like the "static" problem was solved in [prophecy](https://github.com/irontec/prophecy/commit/2d17db5545a52cbfdcddcb7fed0952d8d2dc086f).  Maybe we can try to require `^1.13`?

jordisala1991:
That link comes from a fork instead of the main repo. Seems like it is a fix not applied to the original repo, here is the main issue: https://github.com/phpspec/prophecy/issues/527

garak:
> That link comes from a fork instead of the main repo. Seems like it is a fix not applied to the original repo, here is the main issue: [phpspec/prophecy#527](https://github.com/phpspec/prophecy/issues/527)

Sorry, didn't notice that 😞

jordisala1991:
@garak Can you trigger the build again here? Now it should pass.

garak:
> @garak Can you trigger the build again here? Now it should pass.

It didn't pass

jordisala1991:
Maybe closing and reopening?

jordisala1991:
Needs this fix: https://github.com/KnpLabs/KnpMenuBundle/pull/445/files#diff-8434e9b1d6c83ce1df5b5556356b5b6e2bcc808e635b4a1a23e6f880966c5217R216

jordisala1991:
This one is also needed

PR #447, opened 2021-09-22
Refactor tests to not rely on external mock classes

jordisala1991:
The behavior of tests is the same but now we don't mock what we don't own.
Also it solves the problem with prophecy and php 8 static return type.

wdyt @stof @garak

### DIFF
--- a/tests/DependencyInjection/Compiler/MenuBuilderPassTest.php
+++ b/tests/DependencyInjection/Compiler/MenuBuilderPassTest.php
@@ -4,12 +4,14 @@ namespace Knp\Bundle\MenuBundle\Tests\DependencyInjection\Compiler;
 
 use Knp\Bundle\MenuBundle\DependencyInjection\Compiler\MenuBuilderPass;
 use PHPUnit\Framework\TestCase;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
 
 class MenuBuilderPassTest extends TestCase
 {
+    /**
+     * @var ContainerBuilder
+     */
     private $containerBuilder;
-    private $definition;
-    private $builderDefinition;
 
     /**
      * @var MenuBuilderPass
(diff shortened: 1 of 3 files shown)


garak (approved):

tests/DependencyInjection/Compiler/RegisterMenusPassTest.php
```diff
@@ -21,93 +22,93 @@ class RegisterMenusPassTest extends TestCase
...
+        $menuBuilders = $this->containerBuilder->getDefinition('knp_menu.menu_provider.lazy')->getArgument(0);
+
+        $this->assertFalse($this->containerBuilder->hasDefinition('knp_menu.menu_provider.builder_service'));
+        $this->assertArrayHasKey('foo', $menuBuilders);
+        $this->assertArrayHasKey('bar', $menuBuilders);
+        $this->assertArrayHasKey('baz', $menuBuilders);
+        $this->assertIsArray($menuBuilders['bar']);
+        $this->assertIsArray($menuBuilders['foo']);
+        $this->assertSame('id', (string) $menuBuilders['foo'][0]->getValues()[0]);
+        $this->assertSame('id', (string) $menuBuilders['bar'][0]->getValues()[0]);
```
stof, line 96: There is no assertion that we get a proper ServiceClosureArgument here, but an assumption that `getValues` can be called (and that would also work with an IteratorArgument without making the test fail while the code would be broken).

The equivalent of the old test is `$this->assertEquals($menus, $this->containerBuilder->getDefinition('knp_menu.menu_provider.lazy')->getArgument(0));`, with the existing `$menus` variable

tests/DependencyInjection/Compiler/RegisterMenusPassTest.php
```diff
@@ -21,93 +22,93 @@ class RegisterMenusPassTest extends TestCase
...
-        $this->pass->process($this->containerBuilder->reveal());
+        $this->containerBuilder->getDefinition('menu_id')
+            ->setTags(['knp_menu.menu' => [['alias' => 'foo']]]);
+
+        $this->pass->process($this->containerBuilder);
+
+        $menuBuilders = $this->containerBuilder->getDefinition('knp_menu.menu_provider.lazy')->getArgument(0);
+
+        $this->assertArrayHasKey('foo', $menuBuilders);
+        $this->assertIsNotArray($menuBuilders['foo']);
```
stof, line 111: not being an array does not guarantee it is a ServiceClosureArgument

stof:
I'm all for avoiding to mock the Symfony classes. It also has another benefit: if our compiler passes trigger some deprecations when using the DI component, they will now be triggered when running tests, and so be reported.
However, the assertions can be improved (see my review comments)

tests/DependencyInjection/Compiler/RegisterMenusPassTest.php
jordisala1991: Fixed by checking with the complete array of expected results.

tests/DependencyInjection/Compiler/MenuBuilderPassTest.php
```diff
@@ -18,84 +20,85 @@ class MenuBuilderPassTest extends TestCase
...
+            ->setArgument(0, null)
+            ->setArgument(1, null);
+        $this->containerBuilder->register('id', \stdClass::class)
+            ->addTag('knp_menu.menu_builder', ['alias' => 'foo', 'method' => 'fooMenu'])
+            ->addTag('knp_menu.menu_builder', ['alias' => 'bar', 'method' => 'bar'])
+            ->setPublic(true);
     }
 
+    /**
+     * @doesNotPerformAssertions
```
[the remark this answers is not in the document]
jordisala1991: Fixed by checking the containerBuilder definitions are untouched.

tests/DependencyInjection/Compiler/RegisterMenusPassTest.php
jordisala1991: Fixed by checking with the complete array of expected results.

jordisala1991:
> I'm all for avoiding to mock the Symfony classes. It also has another benefit: if our compiler passes trigger some deprecations when using the DI component, they will now be triggered when running tests, and so be reported.
> However, the assertions can be improved (see my review comments)

I think they should be better now

tests/DependencyInjection/Compiler/RegisterMenusPassTest.php
```diff
@@ -21,93 +24,93 @@ class RegisterMenusPassTest extends TestCase
...
     protected function setUp(): void
     {
-        if (!\class_exists(ServiceClosureArgument::class)) {
-            $this->markTestSkipped('The RegisterMenuPass requires Symfony DI 3.3+.');
-        }
+        $this->containerBuilder = new ContainerBuilder();
+
+        $this->containerBuilder->register('knp_menu.menu_provider.lazy', LazyProvider::class)
+            ->setArgument(0, null);
+        $this->containerBuilder->register('knp_menu.menu_provider.builder_service', \stdClass::class);
```
alexpozzi, line 31: Shouldn't we assert at least once that `knp_menu.menu_provider.builder_service` definition is removed after the process method successfully executes?

alexpozzi (commented):
Thank you for your work!

tests/DependencyInjection/Compiler/RegisterMenusPassTest.php
jordisala1991: Removed by mistake on my previous edit. I will recover that assertion.

jordisala1991: Fixed.

alexpozzi (approved):

jordisala1991:
Friendly ping @stof, can you review again?

stof (approved):

derrabus (approved):

PR #448, opened 2021-10-03
Fix return value for getBundle function in tests

jordisala1991:
To fix #446


jordisala1991:
ping @garak , can we move the Symfony support forward? It would be nice to be able to test it before is out.

PR #451, opened 2022-04-25
increasing the code quality

sewelseb:
I tried it as a training for myself in refactoring.
don't hesitate to reject, but please provide me feedbacks (if you have time)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 .*.cache
-.idea/
 composer.lock
 vendor/*
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .*.cache
+.idea/
 composer.lock
 vendor/*
--- a/src/DependencyInjection/KnpMenuExtension.php
+++ b/src/DependencyInjection/KnpMenuExtension.php
@@ -79,9 +79,19 @@ class KnpMenuExtension extends Extension implements PrependExtensionInterface
     private function GetDefinitionsForEnabledProviders($providers, ContainerBuilder $container): void
     {
         foreach ($providers as $builder => $enabled) {
-            if ($enabled) {
-                $container->getDefinition(\sprintf('knp_menu.menu_provider.%s', $builder))->addTag('knp_menu.provider');
-            }
+            $this->GetDefinitionForEnabledProvider($enabled, $container, $builder);
+        }
+    }
+
+    /**
+     * @param $enabled
+     * @param ContainerBuilder $container
+     * @param $builder
+     */
+    private function GetDefinitionForEnabledProvider($enabled, ContainerBuilder $container, $builder): void
+    {
+        if ($enabled) {
+            $container->getDefinition(\sprintf('knp_menu.menu_provider.%s', $builder))->addTag('knp_menu.provider');
         }
     }
 }
--- a/src/DependencyInjection/KnpMenuExtension.php
+++ b/src/DependencyInjection/KnpMenuExtension.php
@@ -27,10 +27,7 @@ class KnpMenuExtension extends Extension implements PrependExtensionInterface
 
         $this->GetDefinitionsForEnabledProviders($config['providers'], $container);
 
-        if (isset($config['twig'])) {
-            $loader->load('twig.xml');
-            $container->setParameter('knp_menu.renderer.twig.template', $config['twig']['template']);
-        }
+        $config = $this->LoadTwigConfiguration($config, $loader, $container);
         if ($config['templating']) {
             $loader->load('templating.xml');
         }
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
 .*.cache
+.idea/
```
stof, line 2: this is related to your IDE setup, so this should be ignored thanks to file specific to your computer (either only for that project in `.git/info/exclude` or through a system-wide ignore rule). See https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files for the various ways to exclude files in git.

We don't want to maintain exclude rules for all IDEs out there in the project.

src/DependencyInjection/KnpMenuExtension.php
```diff
@@ -79,9 +79,19 @@ public function prepend(ContainerBuilder $container): void
...
+            $this->GetDefinitionForEnabledProvider($enabled, $container, $builder);
+        }
+    }
+
+    /**
+     * @param $enabled
+     * @param ContainerBuilder $container
+     * @param $builder
+     */
+    private function GetDefinitionForEnabledProvider($enabled, ContainerBuilder $container, $builder): void
```
stof: this naming is totally confusing, as it is not getting a definition at all. To me, extracting this method does not even improve code quality. It only makes it harder to see what actually happens.

.gitignore
sewelseb: updated, thanks for the advices :)

src/DependencyInjection/KnpMenuExtension.php
```diff
@@ -75,4 +66,55 @@ public function prepend(ContainerBuilder $container): void
 
         $container->prependExtensionConfig('twig', ['paths' => [$path]]);
     }
+
+    /**
+     * @param $providers
```
garak, line 71: This @param without providing the type is pretty useless. Moreover, the parameter should be type-hinted, and the phpdoc should be present only in case additional information is needed (like for example for arrays)

src/DependencyInjection/KnpMenuExtension.php
```diff
@@ -75,4 +66,55 @@ public function prepend(ContainerBuilder $container): void
...
+     * @param ContainerBuilder $container
+     */
+    private function GetDefinitionsForEnabledProviders($providers, ContainerBuilder $container): void
```
garak, line 74: wrong casing: functions should start with a lowercase letter

src/DependencyInjection/KnpMenuExtension.php
```diff
@@ -75,4 +66,55 @@ public function prepend(ContainerBuilder $container): void
...
+    {
+        foreach ($providers as $builder => $enabled) {
+            $this->GetDefinitionForEnabledProvider($enabled, $container, $builder);
+        }
+    }
+
+    /**
+     * @param $enabled
```
garak, line 82: Same as above about phpdoc

src/DependencyInjection/KnpMenuExtension.php
```diff
@@ -75,4 +66,55 @@ public function prepend(ContainerBuilder $container): void
...
+     */
+    private function getDefinitionForEnabledProvider($enabled, ContainerBuilder $container, $builder): void
+    {
+        if (!$enabled) return ;
+
+        $container->getDefinition(\sprintf('knp_menu.menu_provider.%s', $builder))->addTag('knp_menu.provider');
+    }
+
+    /**
+     * @param array $config
```
garak, line 94: Same as above about phpdoc

src/DependencyInjection/KnpMenuExtension.php
```diff
@@ -75,4 +66,55 @@ public function prepend(ContainerBuilder $container): void
...
+     * @return array
+     * @throws \Exception
+     */
+    private function loadTwigConfiguration(array $config, XmlFileLoader $loader, ContainerBuilder $container): array
+    {
+        if (isset($config['twig'])) {
+            $loader->load('twig.xml');
+            $container->setParameter('knp_menu.renderer.twig.template', $config['twig']['template']);
+        }
+        return $config;
```
garak, line 106: Here you miss an empty line before returning. I guess that you missed running the cs-fixer on your code.

src/DependencyInjection/KnpMenuExtension.php
```diff
@@ -75,4 +66,55 @@ public function prepend(ContainerBuilder $container): void
...
+    }
+
+    /**
+     * @param $templating
```
garak, line 110: Same as above about phpdoc

garak (changes requested):

derrabus:
I don't think that we should merge this. The current code is perfectly readable and I find splitting it into tiny methods not really helpful.

sewelseb:
ok, no issue. As I have said, feel free to reject it.

Just to explain why I like small functions: 
I like to have functions that are shorter than 5 lines of code. The reason is that most of the time if it is longer, it breaks the single responsability principle. When I (a new developper) entered the code, it was difficult for me to understand it. 

Anyway, it was a pleasure to read your comments and feedbacks.

derrabus:
> The reason is that most of the time if it is longer, it breaks the single responsability principle.

This package and especially the class you're fixing here is pure integration code. It's glue between a framework and a library. By design, such a class violates the SRP: You will need to change it whenever the integrated library adds, removes or changes functionality that this package is supposed to integrate. And splitting this class into microscopic methods does not change that: it's still a class with multiple resposibilities. And that's fine.

stof:
Technically, this class has a single responsible: configure the container services based on the user-provided configuration.
The difference is the size of the code needed to implement that responsibility. For such responsibility, 5 LoC is not enough.

Anyway, I'm closing this for reasons already given in comments.

PR #456, opened 2023-05-12
chore: migrate phpunit v9 => v10

Chris53897:
I am unsure about i did this test-replacement correct. Please have a look.
testRender() on https://github.com/KnpLabs/KnpMenuBundle/compare/master...Chris53897:KnpMenuBundle:feature/phpunit-10?expand=1#diff-b293cb7a5c5500f23a6c3f9fa8ff41dc1f2691fce1c7606c78f9b081041defa0R105

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,3 @@
 .*.cache
 composer.lock
-.phpunit.result.cache
-.phpunit.cache
 vendor/*
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 .*.cache
 composer.lock
+.phpunit.result.cache
+.phpunit.cache
 vendor/*
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "symfony/framework-bundle": "^3.4 | ^4.4 | ^5.0 | ^6.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^9.5",
+        "phpunit/phpunit": "^9.6 | ^10.1",
         "symfony/expression-language": "^3.4 | ^4.4 | ^5.0 | ^6.0",
         "symfony/phpunit-bridge": "^6.0",
         "symfony/templating": "^3.4 | ^4.4 | ^5.0 | ^6.0"
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -2,22 +2,18 @@
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          colors="true"
          bootstrap="vendor/autoload.php"
-         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/10.1/phpunit.xsd">
-  <coverage/>
-
-  <testsuites>
-    <testsuite name="KnpMenuBundle Test Suite">
-      <directory suffix="Test.php">./tests/</directory>
-    </testsuite>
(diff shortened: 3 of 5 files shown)


.gitignore
```diff
@@ -1,3 +1,5 @@
 .*.cache
 composer.lock
+.phpunit.result.cache
+.phpunit.cache
```
garak, line 4: these two cases are already covered above. Please remove

composer.json
```diff
@@ -24,7 +24,7 @@
         "symfony/framework-bundle": "^3.4 | ^4.4 | ^5.0 | ^6.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^9.5",
+        "phpunit/phpunit": "^9.6 | ^10.1",
```
garak, line 27: it doesn't make much sense to keep two versions for a dev tool like PHPUnit. Let's drop version 9

phpunit.xml.dist
```diff
@@ -2,18 +2,22 @@
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          colors="true"
          bootstrap="vendor/autoload.php"
-         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.5/phpunit.xsd">
-  <coverage>
+         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/10.1/phpunit.xsd">
+  <coverage/>
```
garak, line 6: I guess we can remove this empty tag, right?

phpunit.xml.dist
```diff
@@ -2,18 +2,22 @@
...
+
+  <testsuites>
```
garak, line 8: Can this tag be kept in the previous position? Just to minimize changes

garak (changes requested):

composer.json
derrabus: PHPUnit 10 requires PHP 8.1. As long as we need to run tests on PHP 7, we need PHPUnit 9. And dropping support for PHP 7 is probably out of scope for this PR?

phpunit.xml.dist
```diff
@@ -2,18 +2,22 @@
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          colors="true"
          bootstrap="vendor/autoload.php"
-         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.5/phpunit.xsd">
-  <coverage>
+         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/10.1/phpunit.xsd">
```
derrabus, line 5: PHPUnit 9 won't understand this configuration file. You need to stick with the (deprecated) format as long as you run tests on PHPUnit 9.

composer.json
garak: good point

phpunit.xml.dist
Chris53897: I have reverted the config-file

phpunit.xml.dist
Chris53897: I have reverted the config-file. (see below)

.gitignore
Chris53897: thanks, i did not see this. reverted

derrabus (approved):

garak (approved):

PR #459, opened 2023-08-06
deprecate templating component

### LINKED ISSUE #458: Deprecate the templating component integration
As of https://github.com/symfony/symfony/pull/51144, the `symfony/templating` component will be deprecated in Symfony 6.4 and removed in 7.0.  The integration in this bundle should be deprecated as well.

garak:
Fix #458

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -64,4 +64,4 @@ jobs:
             - run: composer update --no-interaction --no-progress --ansi ${{ matrix.composer_option }}
             - run: vendor/bin/phpunit
               env:
-                  SYMFONY_DEPRECATIONS_HELPER: max[self]=1
+                  SYMFONY_DEPRECATIONS_HELPER: max[self]=0
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -39,8 +39,6 @@ jobs:
                     - description: 'Lowest deps'
                       php: '7.4'
                       composer_option: '--prefer-lowest'
-                      env:
-                          SYMFONY_DEPRECATIONS_HELPER: max[self]=0
                     - description: 'Symfony 3.4'
                       php: '7.4'
                       symfony: 3.4.*
@@ -67,10 +65,12 @@ jobs:
               with:
                   php-version: ${{ matrix.php }}
             - run: |
-                  sed -ri 's/"symfony\/(.+)": "(.+)"/"symfony\/\1": "'${{ matrix.symfony }}'"/' composer.json;
+                  sed -ri 's/"symfony\/(expression-language|framework-bundle|templating)": "(.+)"/"symfony\/\1": "'${{ matrix.symfony }}'"/' composer.json;
               if: matrix.symfony
             - run: composer config minimum-stability dev
               if: matrix.dev
             - run: composer update --no-interaction --no-progress --ansi ${{ matrix.composer_option }}
             - run: vendor/bin/phpunit
+              env:
+                  SYMFONY_DEPRECATIONS_HELPER: max[self]=1
 
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -63,5 +63,3 @@ jobs:
               if: matrix.dev
             - run: composer update --no-interaction --no-progress --ansi ${{ matrix.composer_option }}
             - run: vendor/bin/phpunit
-              env:
-                  SYMFONY_DEPRECATIONS_HELPER: max[self]=0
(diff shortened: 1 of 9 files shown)


derrabus (approved):

src/Provider/BuilderAliasProvider.php
```diff
@@ -88,7 +88,7 @@ private function getBuilder(string $bundleName, string $className): object
             }
 
             $builder = new $try();
-            if ($builder instanceof ContainerAwareInterface) {
+            if (interface_exists(ContainerAwareInterface::class) && $builder instanceof ContainerAwareInterface) {
```
derrabus, line 91: This change should not be necessary.

garak: > This change should not be necessary.

Are you sure? The interface was removed, so with some Symfony versions doesn't exist.

derrabus: Yes, but an `instanceof` on a non-existing class/interface is safe.

derrabus: Revert the change and see for yourself. The tests should pass. 🙂

garak: You were right. I reverted the change

.github/workflows/build.yaml
```diff
@@ -59,9 +57,11 @@ jobs:
...
             - run: |
-                  sed -ri 's/"symfony\/(.+)": "(.+)"/"symfony\/\1": "'${{ matrix.symfony }}'"/' composer.json;
+                  sed -ri 's/"symfony\/(expression-language|framework-bundle|templating)": "(.+)"/"symfony\/\1": "'${{ matrix.symfony }}'"/' composer.json;
               if: matrix.symfony
             - run: composer config minimum-stability dev
               if: matrix.dev
             - run: composer update --no-interaction --no-progress --ansi ${{ matrix.composer_option }}
             - run: vendor/bin/phpunit
+              env:
+                  SYMFONY_DEPRECATIONS_HELPER: max[self]=1
```
stof, line 67: This is a no-go. Tests covering our own deprecations should use `@group legacy` instead, to make deprecations allowed in them.

tests/Provider/BuilderAliasProviderTest.php
```diff
@@ -5,10 +5,18 @@
...
 use PHPUnit\Framework\TestCase;
-use Symfony\Component\HttpKernel\Bundle\BundleInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
 
 class BuilderAliasProviderTest extends TestCase
 {
+    protected function setUp(): void
+    {
+        parent::setUp();
+        if (!interface_exists(ContainerAwareInterface::class)) {
```
stof, line 15: how is this related to the templating component ?

stof: shouldn't we skip only the test relying on a container-aware menu and not all tests for BuilderAliasProvider ?

And anyway, this should not be in a commit or PR saying it deprecates the templating integration if its fixes a different problem.

.github/workflows/build.yaml
```diff
@@ -59,9 +57,11 @@ jobs:
...
             - run: |
-                  sed -ri 's/"symfony\/(.+)": "(.+)"/"symfony\/\1": "'${{ matrix.symfony }}'"/' composer.json;
+                  sed -ri 's/"symfony\/(expression-language|framework-bundle|templating)": "(.+)"/"symfony\/\1": "'${{ matrix.symfony }}'"/' composer.json;
               if: matrix.symfony
             - run: composer config minimum-stability dev
               if: matrix.dev
             - run: composer update --no-interaction --no-progress --ansi ${{ matrix.composer_option }}
             - run: vendor/bin/phpunit
+              env:
+                  SYMFONY_DEPRECATIONS_HELPER: max[self]=0
```
stof, line 67: This should be reverted. For other jobs that the `Lowest deps` one, we still want deprecations coming from vendors to fail the tests (outside legacy tests). The whole point of using the bridge is that our CI fail when we use deprecated APIs instead of discovering them only when our users complain they get deprecations in their projects.

.github/workflows/build.yaml
garak: I added the needed groups to our legacy tests

tests/Provider/BuilderAliasProviderTest.php
garak: You're right. I changed the skipping, now it skips only a single test

.github/workflows/build.yaml
garak: reverted

stof (approved):

PR #464, opened 2023-10-26
Use Dependency Injection instead of the container

tacman:
Injecting the container no longer works in Symfony 7, so we should remove this from the documentation.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -130,12 +130,13 @@ An example builder class would look like this:
     use App\Entity\Blog;
     use Knp\Menu\FactoryInterface;
     use Knp\Menu\ItemInterface;
-    use Symfony\Component\DependencyInjection\ContainerAwareInterface;
-    use Symfony\Component\DependencyInjection\ContainerAwareTrait;
+    use Doctrine\ORM\EntityManagerInterface;
 
-    final class Builder implements ContainerAwareInterface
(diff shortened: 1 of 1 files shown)


garak (changes requested):
But we still support Symfony 5 and 6 and will probably support them for a couple of years.
Maybe we can just suggest that the container injection is not supported in Symfony 7+

stof:
This example uses `App:Builder:mainMenu` as rendering the menu. this won't work when using DI either.
Btw, this way of referencing menus is deprecated.
To use DI, you need to register your menu builder as a service, which is described in another page of the doc.

We should probably revamp the whole documentation to show the best way first (but the actual best way for modern Symfony projects is not yet implemented as I have an idea of using a `[AsMenuBuilder]` attribute).

tacman:
But in this example, we're only injecting the container in order to get the
Blog posts.  It seems very antiquated to even have this in the
documentation.

tacman:
Is container injection required to use the menu builder?  The intent of my change was to remove the way of getting the repository.  But if that's not possible using DI, I'll close this.

stof:
The syntax with colon-separated aliases as menu names does not support DI at all.

To use DI, we need to register menu builders in a different way, so it would require changing the doc example entirely.

PR #467, opened 2023-10-26
allow Symfony 7

tacman:
It appears that tests pass and the bundle installs.

I'm working on a small Symfony 7 demo but running into unrelated issues.  But I figured I'd get the ball rolling with these small changes.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,9 +38,6 @@
         "psr-4": { "Knp\\Bundle\\MenuBundle\\Tests\\": "tests" }
     },
     "extra": {
-        "symfony": {
-            "require": "7.0"
-        },
         "branch-alias": {
             "dev-master": "3.x-dev"
         }
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,6 @@
 {
     "name": "knplabs/knp-menu-bundle",
+    "minimum-stability": "beta",
     "description": "This bundle provides an integration of the KnpMenu library",
     "keywords": ["menu"],
     "type": "symfony-bundle",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -1,5 +1,6 @@
 {
     "name": "knplabs/knp-menu-bundle",
+    "minimum-stability": "beta",
```
Chris53897, line 3: Do we need this line? Or could this be added to the CI only?

tacman: Probably not, I had that (and the symfony extra) to make sure that tests pass.  How is that added to CI?

Chris53897: One way could be like this. https://github.com/tattali/CalendarBundle/pull/64/files#diff-f47fc96cfa328e123eaec41fe0ef2d23567991431dbdeef87da6c024dd103c08R61

Or a seperate CI-File for dev

garak (approved):

derrabus:
The new CI job fails.

tacman:
Hmm, maybe this line needs to change:

Run sed -ri 's/"symfony\/(expression-language|framework-bundle|templating)": "(.+)"/"symfony\/\1": "'7'"/' composer.json;

is it supposed to be ^7.0?  I'll try that, and I'll try to figure out how to trigger the action from my fork.

tacman:
Thanks, passing now.

tacman:
Fixes #467

derrabus (approved):

trsteel88:
@garak are you able to release a new version for Symfony 7?

PR #460, opened 2023-08-25
chore: add ci tests for php 8.3

Chris53897:

PR #471, opened 2024-02-16
feat: drop support for php 8.0

Chris53897:
migrate phpunit.xml.dist to version 10
allow PHPUnit 11

The change to Attributes, is to fix the deprecation warnings from PHPUnit 11

Maybe it is already time to drop support for php 8.0
https://github.com/KnpLabs/KnpMenuBundle/pull/460#issuecomment-1761147457

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -29,18 +29,16 @@ jobs:
             matrix:
                 include:
                     - description: 'No Symfony specified'
-                      php: '8.0'
-                    - description: 'No Symfony specified'
                       php: '8.1'
                     - description: 'No Symfony specified'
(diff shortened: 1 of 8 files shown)


garak (changes requested):
Please add php 8.3

stof:
Given that KnpMenu still support PHP 8.0, I find it weird to drop it in the bundle (especially with a changelog entry saying it is done to match)

Chris53897:
@garak I just rebased.

@stof Yes of course. I planned to add the corresponding PR for Knp Menu as well. But did not had the time yet.

garak (approved):

derrabus (approved):

Issue #463, opened 2023-10-10
Modernize the directory structure


tacman:
src/Resources/views could be moved to /templates, as that's Symfony recommended best practices.

Chris53897:
@tacman Can you please close this as completed?

PR #473, opened 2024-03-02
feat: use default symfony folder structure

Chris53897:
https://github.com/KnpLabs/KnpMenuBundle/issues/463

https://symfony.com/doc/current/bundles/best_practices.html#directory-structure

I am not sure if there is more todo here.


garak (changes requested):
It doesn't make much sense to me unless we completely remove the "Resources" folder.
So, please move the config as well

garak (approved):

stof:
This PR is incomplete as it does not change the bundle class to tell Symfony that the new structure is used, causing #476

PR #481, opened 2024-06-02
fix deprecated extension

### LINKED ISSUE #480: Deprecated extension
In Symfony 7.1:

>The "Symfony\Component\HttpKernel\DependencyInjection\Extension" class is considered internal since Symfony 7.1, to be deprecated in 8.1; use Symfony\Component\DependencyInjection\Extension\Extension instead. It may change without further notice. You should not use it from "Beelab\Recaptcha2Bundle\DependencyInjection\BeelabRecaptcha2Extension".

In Symfony 5/6 the proper class is already present, so we can safely change our extension point and guarantee the maximum forward compatibility.

garak:
Fix #480


alexpozzi (approved):
Thank you!

PR #487, opened 2024-12-29
remove container-related suggestions from docs

### LINKED ISSUE #465: Revamp the documentation for creating menus
As discussed in #464, we should avoid injecting the container, since that won't be supported in Symfony 7.  Instead, we could start with the menu builder as a service first.

https://symfony.com/bundles/KnpMenuBundle/current/index.html#method-b-a-menu-builder-as-a-service

garak:
Fix #465


alexpozzi (approved):
Thank you!

PR #492, opened 2025-04-30
improve dependencies declaration

### LINKED ISSUE #491: Improve declaration of dependencies
This bundle currently requires `symfony/framework-bundle`. This implies requiring a lot of libraries that are not actually used by the bundle itself.

We should instead be explicit and require the Symfony components we need directly:
* `symfony/config`
* `symfony/dependency-injection`
* `symfony/deprecation-contracts` (already explicit) 
* `symfony/http-kernel`

garak:
Fix #491


alexpozzi (approved):
Thank you!

PR #493, opened 2025-05-07
:sparkles: Add tag "kernel.reset" to matcher to reset cache state between requests

andrew-demb:
https://symfony.com/doc/current/reference/dic_tags.html#kernel-reset


garak (approved):

stof:
thanks

PR #500, opened 2025-10-11
allow Symfony 8, drop Symfony 5

### LINKED ISSUE #499: Support Symfony 8
It's time to add support for Symfony 8

garak:
Fix #499


garak:
> Sigh. Somehow I missed this, and have done some of the same work in my branch. Glad it's done, and hope this PR get merged now that Symfony 8 beta 1 is out.

I'm interested in a possible comparison. Did you end up with the same changes? Does it work for you?

tacman:
The biggest change was the XML to PHP configuration.  Fortunately, AI does a good job with that.

I got stuck with getting tests to pass cs-fixer: 

image

https://github.com/tacman/KnpMenuBundle/actions/runs/18857731525/job/53809488034

garak:
> I got stuck with getting tests to pass cs-fixer:

You need a change in the "native_function_invocation" configuration

tacman:
Maybe a good time to learn how to properly use cs-fixer!

```php
<?php
// see https://github.com/FriendsOfPHP/PHP-CS-Fixer

$finder = PhpCsFixer\Finder::create()
    ->exclude('vendor')
    ->in([__DIR__])
;

return (new PhpCsFixer\Config())
    ->setRiskyAllowed(true)
    ->setRules([
        '@Symfony' => true,
        '@Symfony:risky' => true,
        '@PHP71Migration:risky' => true,
        '@PHPUnit75Migration:risky' => true,
        'ordered_imports' => true,
        'declare_strict_types' => false,
        'native_function_invocation' => ['include' => ['@all']],
    ])
    ->setFinder($finder)
;
```

What should it be in order for tests to pass?

garak:
> What should it be in order for tests to pass?

It's in the diff of this very PR

stof:
FYI, a migration tool for the migration of XML config to PHP configs has been implemented in https://github.com/GromNaN/symfony-config-xml-to-php

stof:
this PR does not just allow Symfony 8. It also drop supports for Symfony 5.4 without even mentioning it.

stof:
@garak I'm not saying we should not drop that. But we must communicate it instead of hiding it in a PR saying it only adds support for Symfony 8 (otherwise, it might end up being missing from the changelog as well)

tacman:
Shouldn't it be ^6.4|^7.3|^8.0?  the older versions aren't supported anymore.

tacman:
So maybe I'm not understanding how Symfony maintains older versions.  Let's say PHP 8.5 has removes something in Symfony 7.*.  Would ALL the version of Symfony 7 get fixed, include 7.0?  My understanding was that it would not, only the maintained versions would get the fix.

Or if Symfony 7.4 changes a signature, won't all the previously maintained versions get the fix as well, but not the unmaintained ones?

https://symfony.com/releases/7.0

I understand that if the bundle does actually support Symfony 7.0 there's no reason to require a higher version.

I guess I'm thinking back to all the bundles that didn't extend AbstractBundle because they wanted to support 6.0 even after it became unmaintained.

dmaicher (approved):

derrabus (approved):

VincentLanglet:
Is there any blocker @garak @stof ?

config/menu.php
```diff
@@ -0,0 +1,113 @@
...
+
+    $services
+        ->set('knp_menu.matcher', Matcher::class)
+        ->public()
+        ->arg(0, new TaggedIteratorArgument('knp_menu.voter'))
+        ->tag('kernel.reset', ['method' => 'clear'])
+    ;
+
+    $services
+        ->set(MenuProviderInterface::class, service('knp_menu.menu_provider'))
```
VincentLanglet: Set signature is
```
set(?string $id, ?string $class = null)
```

I think you should write
```suggestion
        ->alias(MenuProviderInterface::class, 'knp_menu.menu_provider')
```

config/menu.php
```diff
@@ -0,0 +1,113 @@
...
+        ->tag('knp_menu.voter');
+
+    $services
+        ->set('knp_menu.manipulator', MenuManipulator::class)
+    ;
+
+    // Autowiring aliases
+    $services->set(FactoryInterface::class, service('knp_menu.factory'));
+    $services->set(MatcherInterface::class, service('knp_menu.matcher'));
+    $services->set(MenuManipulator::class, service('knp_menu.manipulator'));
```
VincentLanglet:
```suggestion
    $services->alias(FactoryInterface::class, 'knp_menu.factory');
    $services->alias(MatcherInterface::class, 'knp_menu.matcher');
    $services->alias(MenuManipulator::class, 'knp_menu.manipulator');
```

garak: Thanks for the suggestions. The main service now seems to be properly defined, but as soon as I try to use any menu I get a missing reference for the twig service. The service is defined in my project, but (quoting) the container inside "Symfony\Component\DependencyInjection\Argument\ServiceLocator" is a smaller service locator that only knows about the "knp_menu.renderer.list" and "knp_menu.renderer.twig" services."

garak: I'm testing this branch on an actual project, where Twig is present.

config/menu.php
```diff
@@ -0,0 +1,113 @@
...
+            service('kernel'),
+            service('service_container'),
+            service('knp_menu.factory'),
+        ])
+    ;
+
+    $services
+        ->set('knp_menu.renderer_provider', PsrProvider::class)
+        ->args([
+            new ServiceLocatorArgument(new TaggedIteratorArgument('knp_menu.renderer', null, 'alias')),
```
derrabus:
```suggestion
            new ServiceLocatorArgument(new TaggedIteratorArgument('knp_menu.renderer', 'alias')),
```

I think this should fix the locator.

config/menu.php
derrabus: It's the service locator that's broken. See my comment on the menu.php file.

PR #502, opened 2026-02-10
Test on PHP 8.5

garak:


garak:
The failure on CS is fixed in another PR #503

.github/workflows/build.yaml
```diff
@@ -38,36 +40,40 @@ jobs:
...
-                      symfony: ^7.2
-                      dev: true
-                    - description: 'Dev deps'
+                      symfony: ^7.4
+                    - description: 'Symfony 8'
                       php: '8.4'
+                      symfony: ^8.0
+                    - description: 'Dev deps'
+                      php: '8.5'
+                      symfony: ^8.1
```
stof: the job with dev deps should not specify an override of the symfony version range.

.github/workflows/build.yaml
```diff
@@ -38,36 +40,40 @@ jobs:
...
                     - description: 'Symfony 6.4'
                       php: '8.1'
                       symfony: 6.4.*
                     - description: 'Symfony 7'
                       php: '8.2'
-                      symfony: ^7.2
-                      dev: true
-                    - description: 'Dev deps'
+                      symfony: ^7.4
+                    - description: 'Symfony 8'
```
stof: I would remove this job. We need jobs for specific Symfony versions only for the LTS versions

.github/workflows/build.yaml
garak: So it's going to take 8.1 anyway just because it's the latest dev, right?

.github/workflows/build.yaml
garak: But 7.4 is an LTS version, isn't it?

stof: my comment is on the line defining a Symfony 8 job, not on the line defining a job for `7.4`

.github/workflows/build.yaml
stof: yes.
But it will take a newer dev version once available and compatible with our requirements, without the need to remember to update the CI config. Anytime we override the symfony requirement in the CI job, the composer.json is not the single source of truth anymore. This is why we override it only for jobs dedicated at testing compat with Symfony LTS versions (which is the explicit goal of those jobs, and requires maintenance only every 2 years)

